PR #13, opened 2022-05-31
Added Profile page

sprihajha:
- Added a Profile page for users. 
- Added a check for unknown users in `getUserData`
- Removed unnecessary MatchTile styles.

### DIFF
--- a/frontend/src/styles/theme.css
+++ b/frontend/src/styles/theme.css
@@ -123,6 +123,10 @@ hr {
     color: var(--all-chats-page-text-color);
 }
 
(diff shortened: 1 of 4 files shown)


frontend/src/styles/theme.css
```diff
@@ -114,15 +114,6 @@ hr {
     color: var(--all-chats-page-text-color);
 }
 
-.MatchTile {
```
vingkan, line 117: Thanks for catching this!

vingkan (approved):
Requested one change to remove an unnecessary `userProfile &&` check, but after that, looks good to merge!

PR #46, opened 2022-06-03
Remove demo pipeline scripts.

### LINKED ISSUE #25: Delete demo offline pipeline scripts
Now that the pipeline is implemented on Prefect, we can remove the demo scripts.

vingkan:
Closes #25.

These scripts were created to demo the basic matching pipeline. Now that the pipeline has been implemented in Prefect in #8 we can delete these scripts.


vingkan:
Thanks to CodeSee, I noticed that my work on #45 was leaving dangling references in the old demo scripts, so I decided to remove those first so that my other change would be cleaner and not leave any breakages, even in unused code:
 
Dangling reference to scripts.

PR #59, opened 2022-06-04
Remove algorithms module and split out generators, rankers, finalizers.

vingkan:
## Summary

- Create submodules for generators, rankers, finalizers, and engines.
- Move fallback algorithm to finalizers.
- Remove algorithms module.
- Create and test default ranker that does not re-order proposed matches.
- Create and test main matching engine that currently uses only the fallback finalizer.
- Also changes the Gitpod startup to install frontend dependencies using `npm ci`.

## Future Work

- Create and test an example generator.
- Create stubs for generators that the interns will build.
- Create stubs for rankers that the interns will build.
- Refactor the flow to use a matching engine instead of the `compute_fallback_matches()` task.
- Create a `pipeline.transform` module.
- Move the pipeline matching task to the transforms module.
- Move filters and conversions to the transforms module


vingkan:
This provides a more clean view of this refactor:

PR #113, opened 2022-06-14
Enable database access for API.

vingkan:


vingkan:
The production API deployed and ran successfully on Heroku, so we can consider #11 fixed!

Issue #49, opened 2022-06-03
Speed up GitHub action for pre-commit


vingkan:
Right now, the pre-commit action is the slowest GitHub action in the repository. Based on [this example run](https://github.com/scarletstudio/butterfly/runs/6722201496?check_suite_focus=true), installing the frontend dependencies takes 1m27s while running the pre-commit checks themselves only takes 15s.

Can we speed this up? Maybe we only need to install the frontend dev dependencies, instead of all app dependencies.

There may be other actions we can take around speeding up all frontend dependencies, like pruning packages.

PR #126, opened 2022-06-19
Only install ESLint dependencies when running the pre-commit CI action.

vingkan:


vingkan:
I tested that this works correctly! Should give us a big boost for #49, reducing runtime from ~1m26s to ~29s.

Issue #194, opened 2022-06-28
First PR: Ingest Interests


vingkan:
Follow the instructions on our wiki page: [Pipeline Starter Project](https://github.com/scarletstudio/butterfly/wiki/Pipeline-Starter-Project)

Issue #103, opened 2022-06-09
Attributes Ingestion: Ingest Interests


vingkan:
## Goal

Implement a transformation step in the matching pipeline that takes the raw user interest attributes and adds them to the list of interests on the corresponding user object so that the matching engine can use user interests to propose matches.

## Definition of Done

- The transform can be added to the pipeline and runs in production without errors.
- The transform adds the attributes data to user objects in the format the matching engine expects.

## Starter Code

```
-- Implement your transform here
pipeline/transform/user_interests.py
-- Write your unit tests here
pipeline/transform/user_interests_test.py
```

PR #211, opened 2022-06-28
Chris Hunter's work on issue #194 (related to #103).

chunter3:
issue_194_screenshot

### DIFF
--- a/pipeline/transform/user_interests.py
+++ b/pipeline/transform/user_interests.py
@@ -3,16 +3,37 @@ from typing import List
 import prefect
 from prefect import task
 
-from pipeline.types import RawUserInterests, User
+from pipeline.types import Interest, RawUserInterests, User
 
 
 @task
 def augment_users_with_interests(
     users: List[User], raw_interests: RawUserInterests
 ) -> List[User]:
-    # TODO: Implement your transform task
 
     logger = prefect.context.get("logger")
-    logger.info("Chris")
+    logger.info(
+        "Finished issues #103 & #194 (Chris); wrote additional test cases as well"
+    )
+
+    # TODO: Implement your transform task
+
+    for user in users:
+
+        interest_lst = []
+
+        try:
+
+            for key, value in raw_interests[user.uid].items():
+
+                if value:
+
+                    interest_lst.append(Interest(code=key, name=key))
+
+                    user.interests = interest_lst
+
+        except KeyError:
+
+            break
 
     return users
--- a/pipeline/transform/user_interests.py
+++ b/pipeline/transform/user_interests.py
@@ -1,6 +1,5 @@
 from typing import List
 
-import prefect
 from prefect import task
 
 from pipeline.types import Interest, RawUserInterests, User
@@ -11,29 +10,12 @@ def augment_users_with_interests(
     users: List[User], raw_interests: RawUserInterests
 ) -> List[User]:
 
-    logger = prefect.context.get("logger")
-    logger.info(
-        "Finished issues #103 & #194 (Chris); wrote additional test cases as well"
-    )
-
     # TODO: Implement your transform task
 
     for user in users:
-
         interest_lst = []
-
-        try:
-
-            for key, value in raw_interests[user.uid].items():
-
-                if value:
-
-                    interest_lst.append(Interest(code=key, name=key))
-
-                    user.interests = interest_lst
-
-        except KeyError:
-
-            break
-
+        for key, value in raw_interests.get(user.uid, {}).items():
+            if value:
+                interest_lst.append(Interest(code=key, name=key))
+        user.interests = interest_lst
     return users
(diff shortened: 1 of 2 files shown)


pipeline/transform/user_interests.py
```diff
@@ -3,16 +3,37 @@
...
 
 @task
 def augment_users_with_interests(
     users: List[User], raw_interests: RawUserInterests
 ) -> List[User]:
-    # TODO: Implement your transform task
 
     logger = prefect.context.get("logger")
-    logger.info("Chris")
+    logger.info(
```
vingkan, line 15: You can remove the logging entirely now!

In general, logging is helpful when we want to know something is going on in the code when we run the pipeline, for example, logging an unexpected value that doesn't cause a failure.

In this case, I don't think there's anything we need to log.

pipeline/transform/user_interests.py
```diff
@@ -3,16 +3,37 @@
...
+        "Finished issues #103 & #194 (Chris); wrote additional test cases as well"
+    )
+
+    # TODO: Implement your transform task
+
+    for user in users:
+
```
vingkan, line 22: Coding style point: we don't need all these extra blank lines. I use blank lines to break up blocks of code to be more readable, but a blank line after every line is usually too much.

pipeline/transform/user_interests.py
```diff
@@ -3,16 +3,37 @@
...
+                if value:
+
+                    interest_lst.append(Interest(code=key, name=key))
+
+                    user.interests = interest_lst
+
+        except KeyError:
+
+            break
```
vingkan, line 37: If you `break` here, then we will skip all the remaining users. You can use `continue` instead.

pipeline/transform/user_interests.py
```diff
@@ -3,16 +3,37 @@
...
+
+        try:
+
+            for key, value in raw_interests[user.uid].items():
+
+                if value:
+
+                    interest_lst.append(Interest(code=key, name=key))
+
+                    user.interests = interest_lst
```
vingkan, line 33: This overwrites `user.interests` every time the loop runs. Instead, you can do this just once after the loop finishes.

pipeline/transform/user_interests_test.py
```diff
@@ -4,7 +4,7 @@
 from pipeline.types import Interest, RawUserInterests, User
 
 
-def test_example():
+def test_example_1():
```
vingkan, line 7: Can you change this name to describe that the test is checking for? Maybe `test_add_interests`.

pipeline/transform/user_interests_test.py
```diff
@@ -21,15 +21,107 @@ def test_example():
...
+
+def test_example_3():  # This test sees whether a failed assertion is recognized
+
+    users = [
+        User(uid="3", displayName="B"),
+        User(uid="1", displayName="C"),
+        User(uid="2", displayName="A"),
+    ]
+    raw: RawUserInterests = {
+        "1": {},
```
vingkan, line 111: A failed assertion (`actual != expected`) doesn't really help us because we know the output is different, but we don't know whether it is different in the correct way. Try to write assertions as what you expect the output to be, so that there is no doubt.

But this is a good idea to test a user who has no interests data!

Can you also add a case where the user ID does not even appear in the dictionary? That will check your KeyError case.

pipeline/transform/user_interests.py
```diff
@@ -3,16 +3,37 @@
...
+
+    # TODO: Implement your transform task
+
+    for user in users:
+
+        interest_lst = []
+
+        try:
+
+            for key, value in raw_interests[user.uid].items():
```
vingkan, line 27: Optional advice: Instead of using a `try/except` for `KeyError` you can also handle a possibly missing key like this:

```python
for key, value in raw_interests.get(user.uid, {}).item():
```

The `dict.get(key, default)` function lets you get the value of a key from a dictionary if it exists, and if it doesn't exist, return a default value, in this case an empty dictionary.

vingkan (changes requested):
Great work Christopher! I requested some changes and then we can merge this.

pipeline/transform/user_interests.py
```diff
@@ -1,18 +1,21 @@
...
 
 @task
 def augment_users_with_interests(
     users: List[User], raw_interests: RawUserInterests
 ) -> List[User]:
-    # TODO: Implement your transform task
 
-    logger = prefect.context.get("logger")
-    logger.info("Chris")
+    # TODO: Implement your transform task
```
vingkan, line 13: You can remove the TODO now!

pipeline/transform/user_interests_test.py
```diff
@@ -21,15 +21,138 @@ def test_example():
...
         User(uid="3", displayName="C", interests=[climbing, hiking]),
     ]
     # TODO: Uncomment this assertion and delete the users assert
-    # assert actual == expected
-    assert actual == users
+
+    assert actual == expected
 
 
 # TODO: Add more test cases for your logic
```
vingkan, line 35: Remove this TODO as well.

vingkan (approved):
Excellent work @chunter3! Just remove those TODOs and you can merge whenever you are ready.

PR #217, opened 2022-06-29
Verica Karanakova's first commit.

verica-kar:
Screen Shot 2022-06-28 at 8 24 31 PM


vingkan (approved):

vingkan:
Approved @verica-kar! Did the pipeline run in Prefect fail? Wondering why the log screenshot was red.

annalbenson (approved):

verica-kar:
The first two runs failed completely and the most recent one had some 'fails' but mostly 'succeeds'.  Could that be the reason for it being red?

Issue #224, opened 2022-06-30
Matching internal analytics: matches proposed per user


vingkan:
# Summary

To understand and improve our matching algorithms, we want to collect internal analytics about a single matching engine run. Right now, we want to answer basic questions such as:

1. Which users get the most matches generated as choices in a matching engine run?
2. Which generators produce the most matches that make it into production?
3. How often do users not get assigned their top-ranked match because that person was matched to someone else?

This project will start by building the capability to answer the first question, then we can implement more internal matching analytics in the future.

**Note:** Internal matching analytics are different from the matching evaluation and optimization code, because **internal** analytics can have access to all the internals of the matching engine, to consider matches that were not part of the final output.

# Requirements

- R1. Analytics do not disrupt the core matching engine:
  - No change in matching results due to analytics
  - Matching engine core logic does not need to change
- R2. Calculate the number of proposed matches for each user
  - Count a match for a user if that user is in the match
  - Consider only matches after generation and ranking (since ranking can sometimes filter matches out)
  - Consider only matches before finalization (since any user who did not get any match is given a match at this stage)

# Implementation Advice

## 1. Data Interfaces

- Suggestion: create a new dataclass in `pipeline/types/engine.py` called `MatchingInternalAnalytics` that will eventually have one field per metric we want to record, for now just include this one metric we are implementing
- Create a class variable in `MatchingEngine` (`pipeline/matching/core/engine.py`) after the generators/rankers/finalizers for the internal analytics object, now you can save your analytics to this variable without affecting the matching engine output
- In unit tests, you can assert that the internal analytics values on the engine object match the expected values

## 2. Separation of Concerns

- We want the core matching engine logic to be simple to understand, so we want to separate analytics code from core matching code
- Suggestion: write your code in a new file `pipeline/matching/core/analytics.py` so that it can be part of the core engine (internal) but separate from the core logic, then import any functions you need into the core engine code

## 3. Compatibility with Generators
- The generators and rankers do not output a list of matches, they output a generator of matches
- While iterating over a generator, once you take an item, it is no longer in the generator
- This means you can only look at each item from the generator once, unless you store it somewhere else (but that should not be needed for counting)

# Future Work

- Implement more internal analytics metrics that help understand and improve matching algorithms
- Implement ways for developers to view internal analytics after a pipeline run

PR #236, opened 2022-07-04
Internal analytics for matching engine.

chunter3:
successful_prefect_run

No runtime errors occurred during Prefect test run.

### DIFF
--- a/pipeline/matching/core/analytics.py
+++ b/pipeline/matching/core/analytics.py
@@ -1,0 +1,27 @@
+from pipeline.matching.core import MatchingEngine
+from pipeline.types import MatchingInput, MatchingInternalAnalytics
+
+
+class CollectInternalAnalytics:
+    def __init__(self, name, matching_engine):
+        self.name: str = name
+        self.matching_engine: MatchingEngine = matching_engine
+
+    def n_proposed_matches_per_user(
+        self, inp: MatchingInput
+    ) -> MatchingInternalAnalytics:
+
+        n_proposed_matches_per_user = []
+        for user in inp.users:
+            match_count = 0
+            for generator in self.matching_engine.generators:
+                gen_matches = generator.generate(inp)
+                for rank_match in self.matching_engine.ranker.rank(
+                    inp, gen_matches
+                ):
+                    if user.uid in rank_match.users:
+                        match_count += 1
+            n_proposed_matches_per_user.append((user, match_count))
+        return MatchingInternalAnalytics(
+            n_proposed_matches_per_user=n_proposed_matches_per_user
+        )
--- a/pipeline/matching/core/analytics_test.py
+++ b/pipeline/matching/core/analytics_test.py
@@ -1,0 +1,174 @@
+from unittest.mock import MagicMock
+
+import pytest
+
+from pipeline.matching.core.analytics import CollectInternalAnalytics
+from pipeline.matching.core.engine import MatchingEngine
+from pipeline.types import Match, MatchingInput, MatchingInternalAnalytics, User
+
+
+def test_analytics_logic():
+
+    users = [
+        User(uid="A", displayName="User A"),
+        User(uid="B", displayName="User B"),
+        User(uid="C", displayName="User C"),
+        User(uid="D", displayName="User D"),
+        User(uid="E", displayName="User E"),
+    ]
+    recent_matches = [Match(users={"A", "B"}), Match(users={"C", "D", "E"})]
+    inp = MatchingInput(
+        community="test",
+        release="2022-04-01",
+        users=users,
+        recent_matches=recent_matches,
+    )
+    mock_generator = MagicMock()
+    mock_generator.generate = MagicMock(
+        return_value=[
+            Match(users={"A", "B"}),
+            Match(users={"C", "D"}),
+        ]
+    )
+    mock_ranker = MagicMock()
+    mock_ranker.rank = MagicMock(return_value=[Match(users={"A", "D"})])
+    test_matching_engine = MatchingEngine(
+        name="testEngine",
+        generators=[mock_generator],
+        ranker=mock_ranker,
+        finalizer=[],
+    )
+    user_analytics = CollectInternalAnalytics(
+        name="Analytics for number of proposed matches per user",
+        matching_engine=test_matching_engine,
+    )
+
+    MatchingEngine.analytics = user_analytics.n_proposed_matches_per_user(inp)
+
+    expected_analytics = MatchingInternalAnalytics(
+        n_proposed_matches_per_user=[
+            (User(uid="A", displayName="User A"), 1),
+            (User(uid="B", displayName="User B"), 0),
+            (User(uid="C", displayName="User C"), 0),
+            (User(uid="D", displayName="User D"), 1),
+            (User(uid="E", displayName="User E"), 0),
+        ]
+    )
+
+    assert MatchingEngine.analytics == expected_analytics
+
+    # This test varifies the behavior of the internal analyzer w/ a simple case
+
+
+def test_multiple_match_return():
+
+    users = [
+        User(uid="A", displayName="User A"),
+        User(uid="B", displayName="User B"),
+        User(uid="C", displayName="User C"),
+        User(uid="D", displayName="User D"),
+        User(uid="E", displayName="User E"),
+    ]
+    recent_matches = [Match(users={"A", "B"}), Match(users={"C", "D", "E"})]
+    inp = MatchingInput(
+        community="test",
+        release="2022-04-01",
+        users=users,
+        recent_matches=recent_matches,
+    )
+    mock_generator = MagicMock()
+    mock_generator.generate = MagicMock(
+        return_value=[
+            Match(users={"A", "B"}),
+            Match(users={"C", "D"}),
+            Match(users={"A", "D"}),
+        ]
+    )
+    mock_ranker = MagicMock()
+    mock_ranker.rank = MagicMock(
+        return_value=[
+            Match(users={"A", "B"}),
+            Match(users={"A", "C"}),
+            Match(users={"A", "D"}),
+            Match(users={"B", "C"}),
+        ]
+    )
+    test_matching_engine = MatchingEngine(
+        name="testEngine",
+        generators=[mock_generator],
+        ranker=mock_ranker,
+        finalizer=[],
+    )
+    user_analytics = CollectInternalAnalytics(
+        name="Analytics for number of proposed matches per user",
+        matching_engine=test_matching_engine,
+    )
+
+    MatchingEngine.analytics = user_analytics.n_proposed_matches_per_user(inp)
+
+    expected = MatchingInternalAnalytics(
+        n_proposed_matches_per_user=[
+            (User(uid="A", displayName="User A"), 3),
+            (User(uid="B", displayName="User B"), 2),
+            (User(uid="C", displayName="User C"), 2),
+            (User(uid="D", displayName="User D"), 1),
+            (User(uid="E", displayName="User E"), 0),
+        ]
+    )
+
+    assert MatchingEngine.analytics == expected
+
+    # This test varifies the behavior of the internal analyzer when a ranker returns multiple matches for a user
+
+
+def test_no_match_return():
+
+    users = [
+        User(uid="A", displayName="User A"),
+        User(uid="B", displayName="User B"),
+        User(uid="C", displayName="User C"),
+        User(uid="D", displayName="User D"),
+        User(uid="E", displayName="User E"),
+    ]
+    recent_matches = [Match(users={"A", "B"}), Match(users={"C", "D", "E"})]
+    inp = MatchingInput(
+        community="test",
+        release="2022-04-01",
+        users=users,
+        recent_matches=recent_matches,
+    )
+    mock_generator = MagicMock()
+    mock_generator.generate = MagicMock(
+        return_value=[
+            Match(users={"A", "B"}),
+            Match(users={"C", "D"}),
+        ]
+    )
+    mock_ranker = MagicMock()
+    mock_ranker.rank = MagicMock(return_value=[Match(users={})])
+    test_matching_engine = MatchingEngine(
+        name="testEngine",
+        generators=[mock_generator],
+        ranker=mock_ranker,
+        finalizer=[],
+    )
+    user_analytics = CollectInternalAnalytics(
+        name="Analytics for number of proposed matches per user",
+        matching_engine=test_matching_engine,
+    )
+
+    MatchingEngine.analytics = user_analytics.n_proposed_matches_per_user(inp)
+
+    expected_analytics = MatchingInternalAnalytics(
+        n_proposed_matches_per_user=[
+            (User(uid="A", displayName="User A"), 0),
+            (User(uid="B", displayName="User B"), 0),
+            (User(uid="C", displayName="User C"), 0),
+            (User(uid="D", displayName="User D"), 0),
+            (User(uid="E", displayName="User E"), 0),
+        ]
+    )
+
+    assert MatchingEngine.analytics == expected_analytics
+
+    # This test varifies the behavior of the internal analyzer when a ranker returns a match w/ no users
--- a/pipeline/matching/core/engine.py
+++ b/pipeline/matching/core/engine.py
@@ -7,7 +7,13 @@ from pipeline.matching.core.constants import (
 from pipeline.matching.core.finalizer import MatchFinalizer
 from pipeline.matching.core.generator import MatchGenerator
 from pipeline.matching.core.ranker import MatchRanker
-from pipeline.types import Match, MatchingInput, MatchingOutput, UserId
+from pipeline.types import (
+    Match,
+    MatchingInput,
+    MatchingInternalAnalytics,
+    MatchingOutput,
+    UserId,
+)
 
 ERROR_USERS_FOR_PRIORITY_MATCH = "Priority matches should have two users."
 ERROR_USER_NOT_MATCHED = "One user left after selecting priority matches."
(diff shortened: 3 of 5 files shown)


pipeline/types/engine.py
```diff
@@ -48,3 +48,10 @@ class MatchingOutput:
     release: ReleaseTag
     users: List[User]
     matches: List[Match]
+
+
+@dataclass
+class MatchingInternalAnalytics:  # More metrics will be added in the future
+    n_proposed_matches_per_user: List[
+        Tuple[User, int]
+    ]  # The list will be a tuple containing the user and their number of proposed matches
```
vingkan, line 57: What do you think about a dictionary instead of a list of tuples? What are the pros and cons?

pipeline/matching/core/engine.py
```diff
@@ -19,6 +25,9 @@ class UserNotMatchedException(Exception):
 
 
 class MatchingEngine:
+
+    analytics: MatchingInternalAnalytics  # analytics are saved to this variable w/out affecting the matching engine output
```
vingkan, line 29: Move this into the __init__ method, it should be an instance variable instead of a class variable. More about the differences between those two in this article: https://pynative.com/python-class-variables

Also initialize the object.

pipeline/types/engine.py
vingkan, line 57: You can also set a default value for fields in dataclasses.

For a list:

```python
name: Type = field(default_factory=list)
```

For a dictionary:

```python
name: Type = field(default_factory=dict)
```

pipeline/matching/core/analytics.py
```diff
@@ -0,0 +1,27 @@
...
+        n_proposed_matches_per_user = []
+        for user in inp.users:
+            match_count = 0
+            for generator in self.matching_engine.generators:
+                gen_matches = generator.generate(inp)
+                for rank_match in self.matching_engine.ranker.rank(
+                    inp, gen_matches
+                ):
+                    if user.uid in rank_match.users:
+                        match_count += 1
```
vingkan, line 23: instead of doing the actual generation and ranking, this component should only be doing analytics.

So what we want is to take the generator of ranked matches as an input, and then include the analytics in the main engine.

Change the input of this method to take `Iterator[Match]` you can also have the `MatchingInput` as an input, but I don't think we need it for anything. Can you think of anything?

Then we can integrate your analytics class to run after line 41 of the core matching engine code. That way, the matches will already be generated and ranked, you can pass them through your analytics code to collect the counts, and then return the generator of matches untouched and the matching engine code can continue as if nothing happened.

vingkan (changes requested):
Excellent start! The biggest change is that we should make this class something that can be integrated into the matching engine without the class itself calling any generators or rankers, it should only handle counting users, and just pass the ranked matches onto the engine to do the matching.

pipeline/matching/core/analytics.py
vingkan: To be even more clear, this is how the analytics code should integrate into the core matching engine:

```python
    def run(self, inp: MatchingInput) -> MatchingOutput:
        # Generate matches
        proposed: Iterator[Match] = self.generate_matches(inp)
        # Rank matches
        ranked: Iterator[Match] = self.rank_matches(inp, proposed)
        #################
        # This part is where your code goes:
        analyzer = CollectInternalAnalytics
        collected: Iterator[Match] = analyzer.n_proposed_matches_per_user(ranked)
        # Don't forget to change the input on the line below from ranked to collected
        #################
        # Select priority matches and get remaining users
        results = self.select_priority_matches(inp, collected) # Change the input on this line too
        selected: List[Match] = results[0]
        users: Set[UserId] = results[1]
        # Finalize remaining matches
        finalized: List[Match] = self.finalize_matches(inp, selected, users)
```

The reason we want to do it this way is for separation of concerns:
- Analytics code should only be worried about analytics
- Analytics code should not actually run the generators or rankers, just take their output
- We should be able to make some kinds of changes to the logic of the matching engine and still use the analytics code
- Unit tests for analytics code do not need any generators or rankers, just an iterator of matches, they become much simpler to test
- We can also write an integration test that shows how the analyzer and engine work together, that will be a stretch goal, we can pair on that unless you already know how to write it

pipeline/matching/core/analytics.py
```diff
@@ -0,0 +1,27 @@
+from pipeline.matching.core import MatchingEngine
+from pipeline.types import MatchingInput, MatchingInternalAnalytics
+
+
+class CollectInternalAnalytics:
+    def __init__(self, name, matching_engine):
+        self.name: str = name
+        self.matching_engine: MatchingEngine = matching_engine
```
vingkan, line 8: You don't need either of these properties, this can be an empty constructor.

```python
class CollectInternalAnalytics:
    def __init__(self):
        pass
```

pipeline/matching/core/analytics.py
```diff
@@ -0,0 +1,27 @@
...
+
+    def n_proposed_matches_per_user(
+        self, inp: MatchingInput
+    ) -> MatchingInternalAnalytics:
+
+        n_proposed_matches_per_user = []
+        for user in inp.users:
+            match_count = 0
+            for generator in self.matching_engine.generators:
```
vingkan, line 17: What is the runtime of this code? Can it be improved?

pipeline/matching/core/analytics_test.py
```diff
@@ -0,0 +1,174 @@
...
+    )
+    mock_generator = MagicMock()
+    mock_generator.generate = MagicMock(
+        return_value=[
+            Match(users={"A", "B"}),
+            Match(users={"C", "D"}),
+        ]
+    )
+    mock_ranker = MagicMock()
+    mock_ranker.rank = MagicMock(return_value=[Match(users={"A", "D"})])
```
vingkan, line 34: Good job on learning how to use `MagicMock`!

We shouldn't need all this setup, all we need is one iterator of matches as an input, then these tests will be much shorter and easier to understand.

An iterator can be a list, but if you want to really match the input the engine will give you, we should use a generator. You can turn a list into a generator by calling the `iter()` method on it.

For example, this is what your test bodies will look like, much simpler:

```python
matches = [
    Match(users={"A", "B"}),
    Match(users={"C", "D"}),
]
analyzer = CollectInternalAnalytics()
actual = analyzer.n_proposed_matches_per_user(iter(matches))
...
assert actual == expected
```

pipeline/matching/core/analytics.py
```diff
@@ -0,0 +1,27 @@
...
+
+
+class CollectInternalAnalytics:
+    def __init__(self, name, matching_engine):
+        self.name: str = name
+        self.matching_engine: MatchingEngine = matching_engine
+
+    def n_proposed_matches_per_user(
+        self, inp: MatchingInput
+    ) -> MatchingInternalAnalytics:
```
vingkan, line 12: Does this need to be a class with a function or can it just be a function? What do you think are the pros and cons?

pipeline/types/engine.py
chunter3: Now that you mention it, I think a dictionary would work best in this instance. At the time I was more concerned with how I should integrate the user and the associate metric together; and I thought a tuple would be good. However, a dictionary where the user is the key (since all users should be unique anyway) and the desired metric is the value would be more sensible.

pipeline/matching/core/engine.py
chunter3: This has me confused because I was looking over the issue you opened (#224) and in it mentions this: "Create a **class variable** in MatchingEngine (pipeline/matching/core/engine.py) after the generators/rankers/finalizers for the internal analytics object [. . .]." Unless this was a mistake I did create a class variable; but if this should be an instance variable instance, I'll revise accordingly.

pipeline/types/engine.py
chunter3: Yep. I agree. I already revised the dataclass to be a Dict instead of a List. I'll set the default value to be a dict as well.

pipeline/matching/core/analytics.py
chunter3: I see. I was under the impression I wasn't meant to alter the existing matching engine code or its associated functions. But if this is the case, then I see where I went wrong and have a better idea of what I should do going forward. I will revise accordingly.

pipeline/matching/core/analytics.py
chunter3: Well...I was thinking that if we are going to include more metrics in the future, then they would fall under this CollectInternalAnalytics class. While it can be its own function, if we do plan to add more metrics it will help in keeping things more organized. Not to mention using a class is a good way to represent a collection of attributes and methods that may be used repeatedly in other places.

pipeline/matching/core/analytics_test.py
chunter3: Wow...that is very simple. Good to know. Just another instance of me slightly overcomplicating the objective. I will revise my test cases accordingly.

pipeline/matching/core/analytics.py
chunter3: Also don't we need MatchingInput as an input in order to pull all the users? I could be wrong but that was my assumption. I'll add MatchingInput as an input but will revise accordingly if it turns out to be unnecessary.

pipeline/matching/core/engine.py
```diff
@@ -7,6 +7,7 @@
 from pipeline.matching.core.finalizer import MatchFinalizer
 from pipeline.matching.core.generator import MatchGenerator
 from pipeline.matching.core.ranker import MatchRanker
+from pipeline.matching.evaluation.analytics import CollectInternalAnalytics
```
vingkan, line 10: Delete this line since we won't use analytics inside the core engine.

pipeline/matching/core/__init__.py
```diff
@@ -1,3 +1,4 @@
 from pipeline.matching.core.constants import *
 from pipeline.matching.core.engine import *
 from pipeline.matching.core.generator import *
+from pipeline.matching.evaluation.analytics import *
```
vingkan, line 4: Delete this line, we should not cross-import packages in the `__init__.py` file.

pipeline/matching/core/engine_test.py
```diff
@@ -1,5 +1,7 @@
 from unittest.mock import MagicMock
 
+import pytest
+
```
vingkan, line 4: Delete these lines since they are unused.

pipeline/types/engine.py
```diff
@@ -54,3 +54,10 @@ class MatchingOutput:
     users: List[User]
     matches: List[Match]
     internal_data: MatchingInternalData = MatchingInternalData()
+
+
+@dataclass
+class MatchingInternalAnalytics:  # More metrics will be added in the future
+    n_proposed_matches_per_user: Dict = field(
+        default_factory=dict
+    )  # A dictionary where keys -> users and values -> metrics
```
vingkan, line 63: As a coding style point, I think it is easier to read and understand if the comments are on their own line, above the section they refer to. Like this:

```python
@dataclass
class MatchingInternalAnalytics: 
    # A dictionary where keys -> users and values -> metrics
    n_proposed_matches_per_user: Dict = field(
        default_factory=dict
    ) 
    # More metrics will be added in the future
```

vingkan, line 63: Update this comment, the keys are users and the values will be integers.

pipeline/types/engine.py
```diff
@@ -54,3 +54,10 @@ class MatchingOutput:
     users: List[User]
     matches: List[Match]
     internal_data: MatchingInternalData = MatchingInternalData()
+
+
+@dataclass
+class MatchingInternalAnalytics:  # More metrics will be added in the future
```
vingkan, line 60: Now that it's not internal anymore, we can rename this to `MatchingMetrics`, I think metrics is a better description than analytics.

pipeline/matching/evaluation/analytics.py
```diff
@@ -0,0 +1,24 @@
+from typing import Dict, Iterator, List
+
+from pipeline.types import Match, MatchingInput, MatchingInternalAnalytics, User
+
+
+class CollectInternalAnalytics:
```
vingkan, line 6: How about we rename this to `MatchingMetricsCollector`? It is sometimes helpful for the class name to be a noun and the methods to be verbs.

You can rename the method to be `count_proposed_matches_per_user()` to match the verb format, or just leave it as-is.

pipeline/matching/evaluation/analytics.py
```diff
@@ -0,0 +1,24 @@
+from typing import Dict, Iterator, List
```
vingkan, line 1: How about we rename these files to be metrics instead of analytics?

```
pipeline/matching/evaluation/metrics_test.py
pipeline/matching/evaluation/metrics.py
```

pipeline/matching/evaluation/analytics.py
```diff
@@ -0,0 +1,24 @@
...
+        analytics: Dict = {}
+        for match in ranked_matches:
+            for user_id in match.users:
+                if user_id in analytics:
+                    analytics[user_id] += 1
+                    continue
+                analytics[user_id] = 1
+        for user in users:
+            if user.uid not in analytics:
+                analytics[user.uid] = 0
```
vingkan, line 23: Good job! This now runs in O(M + U) where M is the number of matches and U is the number of users. The old code was O(M x U) which is much more inefficient.

pipeline/matching/evaluation/analytics.py
```diff
@@ -0,0 +1,24 @@
...
+
+class CollectInternalAnalytics:
+    def __init__(self):
+        pass
+
+    def n_proposed_matches_per_user(
+        self, users: List[User], ranked_matches: Iterator[Match]
+    ) -> MatchingInternalAnalytics:
+
+        analytics: Dict = {}
```
vingkan, line 14: Have you heard of `defaultdict` before? It is part of the built-in `collections` module with Python. Might make this code cleaner because it allows for a default value when the key is not yet in the dictionary.

Up to you if you want to look it up and use that!

If you use a `defaultdict`, cast it to a regular dict when you return it, that way other keys don't get added when callers try to check invalid users.

pipeline/matching/evaluation/analytics.py
```diff
@@ -0,0 +1,24 @@
...
+        for match in ranked_matches:
+            for user_id in match.users:
+                if user_id in analytics:
+                    analytics[user_id] += 1
+                    continue
+                analytics[user_id] = 1
+        for user in users:
+            if user.uid not in analytics:
+                analytics[user.uid] = 0
+        return MatchingInternalAnalytics(n_proposed_matches_per_user=analytics)
```
vingkan, line 24: We don't need to return the whole analytics object here, just the dict of user IDs to counts.

pipeline/matching/evaluation/analytics_test.py
```diff
@@ -0,0 +1,73 @@
...
+    matches = [Match(users={"A", "B"}), Match(users={"C", "D"})]
+    user_analytics = CollectInternalAnalytics()
+    actual = user_analytics.n_proposed_matches_per_user(users, iter(matches))
+    expected = MatchingInternalAnalytics(
+        n_proposed_matches_per_user={"A": 1, "B": 1, "C": 1, "D": 1, "E": 0}
+    )
+
+    assert actual == expected
+
+    # This test varifies the behavior of the internal analyzer w/ a simple case
```
vingkan, line 26: Typo "verifies", also it is easier for other readers if the comment is at the top of the test:

```python
def test_analytics_logic():
    # This test verifies the behavior of the internal analyzer w/ a simple case
```

pipeline/matching/evaluation/analytics_test.py
```diff
@@ -0,0 +1,73 @@
...
+    matches = [Match(users={})]
+    user_analytics = CollectInternalAnalytics()
+    actual = user_analytics.n_proposed_matches_per_user(users, iter(matches))
+    expected = MatchingInternalAnalytics(
+        n_proposed_matches_per_user={"A": 0, "B": 0, "C": 0, "D": 0, "E": 0}
+    )
+
+    assert actual == expected
+
+    # This test varifies the behavior of the internal analyzer when a ranker returns a match w/ no users
```
vingkan, line 73: Good job on the test cases! I think you captured the import edge cases.

There is an edge case where there is a match for a user who is not in the input, we don't have to worry about that because that would be a problem caused by the generator, not by the metrics. Bad input means bad output :)

pipeline/matching/core/engine.py
vingkan: Yes, I was not careful with my words on the issue. But now it is moot point because we don't need any instance or class variables!

pipeline/types/engine.py
vingkan: Awesome, that sounds good!

pipeline/matching/core/analytics.py
vingkan: After our pairing call, Chris and I realized that it would be much easier to decouple the analytics from the matching engine code, that way they can function separately. The interface between them is that we now return internal data from the matching engine that the analytics can use. Merged in #242.

pipeline/matching/core/analytics.py
vingkan: Chris and I discussed how to improve the runtime! Chris was thinking about the case where some users get no matches and how we want to include a zero count for them, but that can be done more efficiently as well. Good job improving the performance of this, Chris!

pipeline/matching/core/analytics_test.py
vingkan: Excellent, great job refactoring the tests!

pipeline/matching/core/analytics.py
vingkan: I agree with you, let's keep it as a class then!

One thing to think about is merge conflicts, sometimes if multiple people are writing new metrics in the same class, they might create conflicts for each other that slow down changes and reviews. But, we can learn how to handle those.

vingkan (changes requested):
Excellent work Chris! I left all the comments needed for clean-up, once those are done, I will approve and you can merge.

vingkan (approved):
Excellent work, Chris! This is a very clean pull request with efficient code and good tests.

PR #242, opened 2022-07-06
add internal data to matching engine

chunter3:
Outputting extra internal data needed for analytics. It's hidden behind an argument. Set to True if you require internal data.


vingkan:
We checked that the flow still runs successfully!

vingkan (approved):

PR #244, opened 2022-07-06
Similar Intent Matching Capability 

lwilliams717:
added code that passes the base test case

### DIFF
--- a/pipeline/matching/generators/similar_intents.py
+++ b/pipeline/matching/generators/similar_intents.py
@@ -15,4 +15,5 @@ class SimilarIntentsGenerator(MatchGenerator):
 
     def generate(self, inp: MatchingInput) -> Iterator[Match]:
         # TODO: Implement your generator
+        inp.logger.info("Lauren")
         yield from []
--- a/pipeline/matching/generators/similar_intents.py
+++ b/pipeline/matching/generators/similar_intents.py
@@ -1,7 +1,17 @@
 from typing import Iterator
 
+import pytest
+
 from pipeline.matching.core import MatchGenerator
-from pipeline.types import Match, MatchingInput
+from pipeline.types import (
+    Intent,
+    IntentMatch,
+    Match,
+    MatchingInput,
+    MatchMetadata,
+    Side,
+    User,
+)
 
 GENERATOR_SIMILAR_INTENTS = "similarIntentsGenerator"
 
@@ -15,5 +25,65 @@ class SimilarIntentsGenerator(MatchGenerator):
 
     def generate(self, inp: MatchingInput) -> Iterator[Match]:
         # TODO: Implement your generator
-        inp.logger.info("Lauren")
-        yield from []
+        # inp.logger.info("Lauren")
+
+        users = inp.users
+        # print(users)
+        match = False
+
+        match_list = []
+
+        # iterate over the arr of users
+        for idx, u in enumerate(users):
+            # take Intent info from the current user
+            intents = u.intents
+            intent = intents[0]
+            code = intent.code
+            side = intent.side
+            name = intent.name
+
+            # o therSide would be the other side of the intent that they have
+            if side == Side.GIVING:
+                otherSide = Side.SEEKING
+            elif side == Side.SEEKING:
+                otherSide = Side.GIVING
+            else:
+                otherSide = Side.BLANK
+
+            # create an Intent object to search for in the other users
+            otherIntent = Intent(code, otherSide, name)
+
+            for oth in users:
+                if otherIntent in oth.intents:
+                    # create the intent match after finding a user that matches
+                    intent_match = IntentMatch(
+                        code,
+                        seeker=oth.uid
+                        if otherIntent.side == Side.SEEKING
+                        else u.uid,
+                        giver=oth.uid
+                        if otherIntent.side == Side.GIVING
+                        else u.uid,
+                    )
+
+                    # create a Match
+                    new_match = Match(
+                        users={u.uid, oth.uid},
+                        metadata=MatchMetadata(
+                            generator=GENERATOR_SIMILAR_INTENTS,
+                            score=1,
+                            matchingIntents=[intent_match],
+                        ),
+                    )
+
+                    # check if the match already exists before appending it
+                    if new_match in match_list:
+                        continue
+                    else:
+                        match_list.append(new_match)
+
+            # check that a match was found
+            # if so, create a Match and place the two users in the Match
+            # if not, use another intent from the first user and search again
+
+        yield from match_list
--- a/pipeline/matching/generators/similar_intents.py
+++ b/pipeline/matching/generators/similar_intents.py
@@ -1,4 +1,4 @@
-from typing import Iterator
+from typing import Dict, Iterator
 
 import pytest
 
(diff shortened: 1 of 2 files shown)


pipeline/matching/generators/similar_intents.py
```diff
@@ -15,4 +25,65 @@ def __init__(self, min_common: int = 0):
...
+        # print(users)
+        match = False
+
+        match_list = []
+
+        # iterate over the arr of users
+        for idx, u in enumerate(users):
+            # take Intent info from the current user
+            intents = u.intents
+            intent = intents[0]
```
arushirai1, line 40: Although this type of assumption about the data structure could lead to errors -- in this case I think it cancels out due to symmetric property of giving and seeking intents. Specifically in line 56, you end up looping through all other users intents.

pipeline/matching/generators/similar_intents.py
```diff
@@ -15,4 +25,65 @@ def __init__(self, min_common: int = 0):
...
+                        metadata=MatchMetadata(
+                            generator=GENERATOR_SIMILAR_INTENTS,
+                            score=1,
+                            matchingIntents=[intent_match],
+                        ),
+                    )
+
+                    # check if the match already exists before appending it
+                    if new_match in match_list:
+                        continue
```
arushirai1, line 81: Following up with an earlier comment, perhaps you can add to the matchingIntents attribute of the match if you find a match that includes the same users. You would need to modify the type of match_list to do this efficiently. What data structure would be best suited for this task?

pipeline/matching/generators/similar_intents_test.py
```diff
@@ -22,7 +22,7 @@ def test_example():
         release="2022-06-26",
         users=[
             User(uid="1", displayName="A", intents=[seek_tutoring]),
-            User(uid="2", displayName="B", interests=[give_tutoring]),
+            User(uid="2", displayName="B", intents=[give_tutoring]),
```
arushirai1, line 25: Good job catching this error!

pipeline/matching/generators/similar_intents.py
```diff
@@ -15,4 +25,65 @@ def __init__(self, min_common: int = 0):
...
+            side = intent.side
+            name = intent.name
+
+            # o therSide would be the other side of the intent that they have
+            if side == Side.GIVING:
+                otherSide = Side.SEEKING
+            elif side == Side.SEEKING:
+                otherSide = Side.GIVING
+            else:
+                otherSide = Side.BLANK
```
vingkan, line 51: This is a good idea! Makes it easier to find the compatible intent.

Two thoughts to make this code simpler:

1. You can assume that the intent side will be either GIVING or SEEKING. If it is not, that is a problem that occurs before your code. If you want, you could raise a ValueError if you encounter a side that is not one of those values, but this is optional.

2. Since we can assume the side will be either SEEKING or GIVING, you could simplify this block of code with a [ternary statement](https://book.pythontips.com/en/latest/ternary_operators.html) (like the ones you used below):

```python
other_side = Side.GIVING if side == Side.SEEKING else Side.SEEKING
```

pipeline/matching/generators/similar_intents.py
```diff
@@ -15,4 +25,65 @@ def __init__(self, min_common: int = 0):
...
+                        if otherIntent.side == Side.GIVING
+                        else u.uid,
+                    )
+
+                    # create a Match
+                    new_match = Match(
+                        users={u.uid, oth.uid},
+                        metadata=MatchMetadata(
+                            generator=GENERATOR_SIMILAR_INTENTS,
+                            score=1,
```
vingkan, line 74: You can exclude score and then it will get 0 as its default value. You can check if a dataclass has default values by looking at the definition, for example: https://github.com/scarletstudio/butterfly/blob/20edebb83c443a7358709f4966190e7b5e0aa3ca/pipeline/types/match_metadata.py#L35

pipeline/matching/generators/similar_intents.py
```diff
@@ -15,4 +25,65 @@ def __init__(self, min_common: int = 0):
...
+                if otherIntent in oth.intents:
+                    # create the intent match after finding a user that matches
+                    intent_match = IntentMatch(
+                        code,
+                        seeker=oth.uid
+                        if otherIntent.side == Side.SEEKING
+                        else u.uid,
+                        giver=oth.uid
+                        if otherIntent.side == Side.GIVING
+                        else u.uid,
```
vingkan, line 66: It might be more readable to move these ternary statements into variables, rather than assigning them inline in the IntentMatch constructor.

pipeline/matching/generators/similar_intents.py
```diff
@@ -1,7 +1,17 @@
...
+    Intent,
+    IntentMatch,
+    Match,
+    MatchingInput,
+    MatchMetadata,
+    Side,
+    User,
+)
 
 GENERATOR_SIMILAR_INTENTS = "similarIntentsGenerator"
```
vingkan, line 17: Maybe we should change the default value of `min_common: int = 0` to `1`, otherwise we will generate matches for users with no compatible intents.

vingkan (changes requested):
This is great progress, Lauren! Keep up the good work.

pipeline/matching/generators/similar_intents.py
lwilliams717: > I think we still want to loop over all the intents for both users. Can you create a test case that shows how this code could fail?

@vingkan Just added a test case when users have more than one intent, and I've worked with @arushirai1 to loop through all the intents. Since I loop through all users in the list, and I loop through every users intent, would this implementation handle that situation?

pipeline/matching/generators/similar_intents.py
```diff
@@ -15,4 +25,65 @@ def __init__(self, min_common: int = 0):
...
+                        else u.uid,
+                    )
+
+                    # create a Match
+                    new_match = Match(
+                        users={u.uid, oth.uid},
+                        metadata=MatchMetadata(
+                            generator=GENERATOR_SIMILAR_INTENTS,
+                            score=1,
+                            matchingIntents=[intent_match],
```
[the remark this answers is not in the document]
lwilliams717: Got it! Thanks for mentioning this since I didn't notice this earlier.

pipeline/matching/generators/similar_intents.py
lwilliams717: > Following up with an earlier comment, perhaps you can add to the matchingIntents attribute of the match if you find a match that includes the same users. You would need to modify the type of match_list to do this efficiently. What data structure would be best suited for this task?

As discussed in our paired programming session, a dictionary (hash) seems like the best structure for the task. The dictionary contains both users id and the Match object as the value. The ease of lookup makes it easy to see which users have been matched and the matchingIntents can be appended to if users match on a different intent.

pipeline/matching/generators/similar_intents.py
lwilliams717: > This is a good idea! Makes it easier to find the compatible intent.
> 
> Two thoughts to make this code simpler:
> 
> 1. You can assume that the intent side will be either GIVING or SEEKING. If it is not, that is a problem that occurs before your code. If you want, you could raise a ValueError if you encounter a side that is not one of those values, but this is optional.

If I raise a ValueError, should this be something the Ranking Team might need to know for their implementation? Also, would it stop the generator from returning anything?

> 2. Since we can assume the side will be either SEEKING or GIVING, you could simplify this block of code with a [ternary statement](https://book.pythontips.com/en/latest/ternary_operators.html) (like the ones you used below):
> 
> ```python
> other_side = Side.GIVING if side == Side.SEEKING else Side.SEEKING
> ```
Thanks for the suggestion!

pipeline/matching/generators/similar_intents.py
lwilliams717: Got it. thanks!

lwilliams717: > Maybe we should change the default value of `min_common: int = 0` to `1`, otherwise we will generate matches for users with no compatible intents.
Resolved on line 24

pipeline/matching/generators/similar_intents.py
lwilliams717: > I think we still want to loop over all the intents for both users. Can you create a test case that shows how this code could fail?

resolved on line 41 of similar_intent.py

pipeline/matching/generators/similar_intents.py
lwilliams717: > 2. Since we can assume the side will be either SEEKING or GIVING, you could simplify this block of code with a [ternary statement](https://book.pythontips.com/en/latest/ternary_operators.html) (like the ones you used below):
> 
> ```python
> other_side = Side.GIVING if side == Side.SEEKING else Side.SEEKING
> ```

Resolved on line 47 of similar_intents.py

pipeline/matching/generators/similar_intents_test.py
```diff
@@ -43,8 +43,131 @@ def test_example():
...
+    )
+
+    generator = SimilarIntentsGenerator(min_common=1)
+    actual3 = list(generator.generate(inp3))
+
+    assert actual3 == []
+
+
+# test case where users match on more than one intent
+def test_mult_match_intent():
```
arushirai1, line 116: Not a problem for this PR, but depending on the situation you might want to decouple multiple test scenarios into separate test cases if possible. Of course if multiple specific input conditions are required for the test situation then they should be kept in the same test case.

arushirai1 (approved):
Approving for now -- the only critical change that's needed is to only include matches between two distinct useres (see comments on where to handle this) since a single user could have both giving and seeking sides for the same intent. You can open a separate PR after this change.

The other changes are more for readability purposes; you can parse through those comments. There is one other issue which is incorporating the min_common attribute.

pipeline/matching/generators/similar_intents.py
vingkan: Good question, the ranker can add the score, or the generator can provide it. I have a slight preference for the ranker adding the score, that way the ranker controls scoring and thus ranking.

We can leave the score blank for now, this could be a new issue to add a score for these matches.

vingkan (approved):
Looks like all my comments have been resolved!

PR #246, opened 2022-07-07
Core project: Variety Rankers

dfg-13:


pipeline/matching/rankers/variety.py
```diff
@@ -1,17 +1,31 @@
...
 
     def rank(
         self, inp: MatchingInput, proposed: Iterator[Match]
     ) -> Iterator[Match]:
-        # TODO: Implement your ranker
-        yield from proposed
+        inp.logger.info("David was here!")
+        yield from sorted(
+            proposed, key=lambda match: dup_gencheck(inp.recent_matches, match)
+        )
```
notedwin, line 31: Generators for matches behaved a little differently than I expected and tested for, so every time we look at a generator the value gets taken out of the generator.

However, we are only looking at one specific match at a time and the MatchingInput, which is not a generator so we are good! 

Reference: **https://github.com/scarletstudio/butterfly/pull/248/files#r917440831**

pipeline/matching/rankers/variety_test.py
```diff
@@ -1,6 +1,6 @@
 import pytest
 
-from pipeline.matching.rankers import VarietyRanker
+from pipeline.matching.rankers import VarietyRanker, variety
```
notedwin, line 3: Importing is subjective. This is fine with me, as I don't think we will be adding anything complex to the variety.py file. 

For more complex projects, I would advise against importing the whole file and importing the specific function/class.

For example: 
```
from pipeline.matching.rankers import VarietyRanker, dep_gencheck
```

pipeline/matching/rankers/variety_test.py
```diff
@@ -26,9 +26,85 @@ def test_example():
...
+    actual = list(ranker.rank(inp, proposed))
+
+    expected = [
+        Match(users={"A", "D"}, metadata=metadata_2),
+        Match(users={"E", "F"}, metadata=metadata_1),
+        Match(users={"G", "F"}, metadata=metadata_1),
+        Match(users={"A", "C"}, metadata=metadata_1),  # expected to be lower
+    ]
+
+    assert actual == expected
```
notedwin, line 110: These tests and comments are great!  There are some tests we could add for 3 users, or where all matches are considered duplicates, but they aren’t essential!

notedwin:
The changes look good to me, resolve the conflicts and then I will accept!

notedwin (approved):

PR #248, opened 2022-07-07
Nelida Core Project Quantity Ranker

nelischalich:


pipeline/matching/rankers/quantity.py
```diff
@@ -6,12 +6,30 @@
...
 
     def rank(
         self, inp: MatchingInput, proposed: Iterator[Match]
     ) -> Iterator[Match]:
-        # TODO: Implement your ranker
-        yield from proposed
+        """The function 'rank' returns a sorted iterator based on the total number of matches for each user.
+        The less matches a user has the higher they are prioritized."""
+        num_matches = get_number_matches(proposed)
+        yield from sorted(
```
vingkan, line 30: The input to the ranking function will be a generator of matches. One gotcha with generators is that when you loop through the values, they are taken out of the generator, so the second loop will find an empty generator.

Read more about this in the [intern handbook Python tips](https://github.com/scarletstudio/handbook/wiki/Python-Tips#generators).

For now, we can handle it by doing this:

```python
# Copy generator into a list and use that list
proposed_list = list(proposed)
num_matches = get_number_matches(proposed_list)
yield from sorted(proposed_list, ...)
```

In the future, this could become memory-inefficient. Imagine we have generators that produce every possible match. If there are N users, that means N * (N - 1) matches per generator. If N is a big number, we will have N^2 matches being copied into memory, rather than being read one-by-one from a generator. At our current number of users, this is no problem, and probably won't be a big deal if communities on Butterfly are small (~100 users). But imagine if we had 1000 users and 10 generators that each produce every option, we could have 10 million matches being copied into memory!

As a challenge, you could come back to this in the future and find a way to do this without copying the generator, but that would be more for personal learning rather than an urgent need for the system.

pipeline/matching/rankers/quantity_test.py
```diff
@@ -27,9 +27,58 @@ def test_example():
...
+    )
+    proposed = [
+        Match(users={"A", "B", "C"}),  # total num of matches is 5
+        Match(users={"A", "C", "E"}),  # total is 6
+        Match(users={"E", "D", "F"}),  # total is 4
+        Match(users={"X", "Y", "Z"}),  # total is 3
+    ]
+    # A:2 C:2 E:2 BDFXYZ:1
+    ranker = QuantityRanker()
+    actual = list(ranker.rank(inp, proposed))
```
vingkan, line 48: Per the previous comment about generators having values taken out, you can simulate that in your unit test by casting the list of proposed matches to `iter()`. After that, this test should fail and you can try to fix the issue.

```python
actual = list(ranker.rank(inp, iter(proposed)))
```

pipeline/matching/rankers/quantity.py
```diff
@@ -6,12 +6,30 @@
...
+            proposed,
+            key=lambda match: sum(
+                [num_matches.get(user) for user in match.users]
+            ),
```
vingkan, line 34: This is cool! At first, I was thinking we would rank matches by the user in that match with the fewest number of options, not by the sum of the number of matches for all the users in the match.

I realized that your sum approach will work as well, so let's stick with this!

In the future, when we combine this with other rankers, it might be better to use the minimum number of options rather than the sum of options, that way other rankers can have an effect on which match to choose for a user with few options, rather than just matching them to the user with the fewest options. But we can investigate this later when there are more generators and rankers in production.

vingkan (changes requested):
Excellent work @nelischalich ! I left some comments for you and @notedwin , let me know if you have any questions.

pipeline/matching/rankers/quantity.py
```diff
@@ -6,12 +6,30 @@
 RANKER_QUANTITY = "quantityRanker"
 
 
+def get_number_matches(matches):
+    num_matches = {}
+    for match in matches:
+        for user in match.users:
+            if num_matches.get(user) is None:
```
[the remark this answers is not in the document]
notedwin: The Match type uses a set of userid as users and not user objects. Changing the name of the variable would make that clear.

pipeline/matching/rankers/quantity.py
```diff
@@ -6,12 +6,30 @@
...
+                num_matches[user] = 1
+            else:
+                num_matches[user] += 1
```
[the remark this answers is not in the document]
nelischalich: Implemented defaultdict

pipeline/matching/rankers/quantity.py
nelischalich: Implemented iterator into a list, however it is memory-inefficient.

pipeline/matching/rankers/quantity_test.py
nelischalich: Typecasted proposed to an iterable.

pipeline/matching/rankers/quantity.py
nelischalich: Sticking with current implementation, keeping this approach in mind if a user with a low number of matches is getting low quality matches.

notedwin:
Screenshot 2022-07-18 at 7 27 02 PM

notedwin:
Successful local run on SDS, hopefully, nothing like #281 happens

notedwin (approved):

pipeline/matching/rankers/quantity.py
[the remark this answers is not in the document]
vingkan: Good catch! I missed that.

vingkan (approved):
Great work, Nelida!

pipeline/matching/rankers/quantity_test.py
```diff
@@ -19,17 +19,74 @@ def test_example():
...
+        Match(users={"E", "B"}),
+    ]
+
+    """expected = [
+        Match(users={"E", "C"}),
+        Match(users={"X", "B"}),
+        Match(users={"Z", "B"}),
+        Match(users={"E", "B"}),
+        Match(users={"A", "B", "C"}),
+    ]"""
```
vingkan, line 91: You could delete this.

PR #249, opened 2022-07-08
Adds Intents to User Objects

### LINKED ISSUE #104: Attributes Ingestion: Ingest Intents
## Goal

Implement a transformation step in the matching pipeline that takes the raw user intents attributes and adds them to the list of intent on the corresponding user object so that the matching engine can use user intents to propose matches.

## Definition of Done

- The transform can be added to the pipeline and runs in production without errors.
- The transform adds the attributes data to user objects in the format the matching engine expects.

## Starter Code

```
-- Implement your transform here
pipeline/transform/user_intents.py
-- Write your unit tests here
pipeline/transform/user_intents_test.py
```

azuhdz7:
Closes #104

### DIFF
--- a/pipeline/transform/user_intents_test.py
+++ b/pipeline/transform/user_intents_test.py
@@ -23,4 +23,18 @@ def test_example():
     assert actual == expected
 
 
-# TODO: Add more test cases for your logic
+def test_user_no_intents():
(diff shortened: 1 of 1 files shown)


vingkan (approved):
Looks good to merge!

PR #254, opened 2022-07-11
Mef first pr

mflowers1395:


frontend/src/app/attributes/EditIntents.tsx
```diff
@@ -1,14 +1,85 @@
...
+                <div key={intent.code}>
+                    <Subjects topic={intent.name} />
+                    <Seeking
+                        value={intent.isSeeking}
+                        updateIntent={updateIntent}
+                        code={intent.code}
+                    />
+
+                    <Giving
+                        value={intent.isSeeking}
```
vingkan, line 75: This should be `isGiving` instead of `isSeeking`.

frontend/src/app/attributes/EditIntents.tsx
```diff
@@ -1,14 +1,85 @@
...
-    return <p> Marilyn Elaine Flowers was here! </p>
+const EditIntents = ({ intents = [], userIntentMap = {}, updateIntent }: EditIntentsProps) => {
+    const userIntents = intents.map((intent) => transformUserIntents(intent, userIntentMap))
+    console.log(userIntents)
+    return (
+        <div>
+            {userIntents.map((intent) => (
+                <div key={intent.code}>
+                    <Subjects topic={intent.name} />
+                    <Seeking
```
vingkan, line 68: The code for `Seeking` and `Giving` are very similar. Can you create a single component called `IntentCheckbox` and use it for both?

Here is what the `EditIntents` component should look like if you do that:

```jsx
<Subjects topic={intent.name} />
<IntentCheckbox
  side="seeking"
  code={intent.code}
  value={intent.isSeeking}
  updateIntent={updateIntent}
/>
<IntentCheckbox
  side="giving"
  code={intent.code}
  value={intent.isGiving}
  updateIntent={updateIntent}
/>
```

Notice how we use the same component twice, one for each side. And we add a new prop called `side` that will allow the component to use that variable instead of hard-coding giving or seeking.

vingkan (changes requested):
Looking good! And tests are passing 🎉 

Once you fix the comment about using `IntentCheckbox` instead of `Seeking` and `Giving`, this will be ready to merge.

vingkan (approved):
Great work, Marilyn!

PR #255, opened 2022-07-11
Encountering issues w/ integration; requesting review

chunter3:
Specifically an AttributeError. Requesting review to resolve errors and verify the displaying of metrics is properly integrated into the matching flow.

### DIFF
--- a/pipeline/flows/matching.py
+++ b/pipeline/flows/matching.py
@@ -18,6 +18,7 @@ from pipeline.extract import (
     extract_recent_matches,
     extract_users,
 )
+from pipeline.load.display_metrics import display_internal_matching_metrics
 from pipeline.load.release import delete_previous_release, upload_matches
 from pipeline.load.validation import validate_and_log_matches
 from pipeline.matching.core import MatchingEngine
@@ -114,6 +115,8 @@ def matching_flow(defaults: Dict = {}) -> Flow:
         passed = validate_and_log_matches(
             users_w_profile, recent_matches, output_matches
         )
+
+        display_internal_matching_metrics.run(users_w_data, recent_matches)
 
         # Do not prepare output matches for loading unless validation passed
         df_output_matches = transform_matches_for_load(
--- a/pipeline/load/display_metrics.py
+++ b/pipeline/load/display_metrics.py
@@ -1,0 +1,19 @@
+from typing import Iterator, List
+
+import prefect
+from prefect import task
+
+from pipeline.transform import compute_internal_matching_metrics
+from pipeline.types import Match, MatchingOutput, User
+
+
+@task
+def display_internal_matching_metrics(
+    users: List[User], ranked_matches: Iterator[Match]
+):
+    logger = prefect.context.get("logger")
+    metrics_output = compute_internal_matching_metrics(users, ranked_matches)
+    logger.info("Displaying number of proposed matches per user")
+    print(f"Matches Proposed per User:\n")
+    for uid, count in metrics_output.n_proposed_matches_per_user.items():
+        print(f"{count} - User {uid}\n")
--- a/pipeline/matching/evaluation/__init__.py
+++ b/pipeline/matching/evaluation/__init__.py
@@ -1,0 +1,1 @@
+from pipeline.matching.evaluation.metrics import *
--- a/pipeline/transform/__init__.py
+++ b/pipeline/transform/__init__.py
@@ -1,3 +1,4 @@
+from pipeline.transform.compute_metrics import *
 from pipeline.transform.matches import *
 from pipeline.transform.user_intents import *
 from pipeline.transform.user_interests import *
--- a/pipeline/transform/compute_metrics.py
+++ b/pipeline/transform/compute_metrics.py
@@ -1,0 +1,22 @@
+from typing import Iterator, List
+
+import prefect
+from prefect import task
+
+from pipeline.matching.evaluation import MatchingMetricsCollector
+from pipeline.types import Match, MatchingMetrics, MatchingOutput, User
+
+
+@task
+def compute_internal_matching_metrics(
+    users: List[User], ranked_matches: Iterator[Match]
+) -> MatchingMetrics:
+    logger = prefect.context.get("logger")
+    metrics = MatchingMetricsCollector()
+    n_proposed_matches_per_user = metrics.count_proposed_matches_per_user(
+        users, ranked_matches
+    )
+    logger.info("Generating internal matching metrics")
+    return MatchingMetrics(
+        n_proposed_matches_per_user=n_proposed_matches_per_user
+    )


pipeline/flows/matching.py
```diff
@@ -115,6 +116,8 @@ def matching_flow(defaults: Dict = {}) -> Flow:
             users_w_profile, recent_matches, output_matches
         )
 
+        display_internal_matching_metrics.run(users_w_data, recent_matches)
```
vingkan, line 119: Ah, I probably confused you with my note.

If we are running a Prefect task in normal Python code, then we use `task.run()`.

BUT, when a Prefect task runs in a flow context (which is what is happening in this file), you can use it as if it were a function, and Prefect will automatically handle it.

On line 56 of this file, we have:

```python
with Flow(name="matching_flow") as flow:
```

And all the code that is intended under that is part of the flow context.

So instead, you can change your line to:

```python
display_internal_matching_metrics(users_w_data, ranked_matches)
```

Also, don't forget that you need to pass in ranked_matches, not recent matches.

I would recommend changing this to just use the matching output as an argument instead of the individual variables. But now that I look at the code, I realize that `compute_matches()` only returns the output matches, not the full output.

To get around that, try making the following changes:

```python
# On line 39 of the matching flow code, return both the output and the matches:
@prefect.task
def compute_matches(inp: MatchingInput, engine_id: EngineId) -> List[Match]:
    ...
    return output, output.matches

# On line 113 of the matching flow code, I'm not sure if this will work, but try this and let me know what happens:
matching_output, output_matches = compute_matches(inp, param_engine)

# Then, on line 119 of the flow code, you can do this:
display_internal_matching_metrics(matching_output)

# And finally, update your display task in the load module:
@task
def display_internal_matching_metrics(output: MatchingOutput):
  users = output.users
  ranked_matches = output.internal_data.ranked_matches
  ...
```

The `run flow matching` to see if it works.

pipeline/flows/matching.py
```diff
@@ -18,6 +18,7 @@
     extract_recent_matches,
     extract_users,
 )
+from pipeline.load.display_metrics import display_internal_matching_metrics
```
vingkan, line 21: Don't forget to import the compute metrics transform, you need to compute the metrics before you can display them.

pipeline/load/display_metrics.py
```diff
@@ -0,0 +1,19 @@
...
+from pipeline.transform import compute_internal_matching_metrics
+from pipeline.types import Match, MatchingOutput, User
+
+
+@task
+def display_internal_matching_metrics(
+    users: List[User], ranked_matches: Iterator[Match]
+):
+    logger = prefect.context.get("logger")
+    metrics_output = compute_internal_matching_metrics(users, ranked_matches)
```
vingkan, line 15: Ah, I see that you called this inside the display task.

Instead, I want you to call the compute task in the flow code, then pass the MatchingMetrics output class to your display function.

chunter3:
result in prefect log

This is the result in the Prefect logs. I'm satisfied with the structure of the output. Before I was outputting each user-metric pair as its own log. Now the entire result is contained in a single log. This should help cutback on clutter within the logs.

pipeline/load/display_metrics.py
```diff
@@ -0,0 +1,20 @@
...
+
+@task
+def display_internal_matching_metrics(metric: MatchingMetrics):
+    logger = prefect.context.get("logger")
+    logger.info("Displaying number of proposed matches per user")
+    logger.info(
+        "Matches Proposed per User:\n"
+        + "\n".join(
+            [
+                f"{count} - User {uid}"
```
vingkan, line 16: We want to print the user name, not the uid. When I wrote `User A` in the issue, I was thinking that was their name, but I wasn't clear.

So maybe you need to add users as an input to this function, use a dictionary lookup to get the name for the uid, then print that. Give that a try.

pipeline/matching/core/engine.py
```diff
@@ -26,7 +26,7 @@ class UserNotMatchedException(Exception):
 
 class MatchingEngine:
     def __init__(
-        self, name, generators, ranker, finalizer, return_internal_data=False
+        self, name, generators, ranker, finalizer, return_internal_data=True
```
vingkan, line 29: This is causing tests to fail, instead of changing the default value to True, we should only set it to True when we run the pipeline.

Change the default value here back to False.

Then, go to this line in the matching flow code: https://github.com/scarletstudio/butterfly/blob/main/pipeline/flows/matching.py#L47

And change it to have:

```python
engine: MatchingEngine = Engine(return_internal_data=True)
```

That will achieve your goal of turning on internal data during the pipeline run, but leaving it defaulted to False so that the test cases pass.

vingkan (changes requested):
Looking good! I requested two changes:
1. Fix the broken tests by changing how we set `return_internal_data`
2. Show the user display name instead of the uid in the displayed metrics

pipeline/matching/core/engine.py
chunter3: I'm getting a TypeError after making the change you're suggesting here. However, if I change it to this instead:

engine: MatchingEngine(return_internal_data=True) = Engine()

Everything runs fine. For the time being, I'll go with this.

chunter3:
Successful Prefect Run

Prefect ran without issue. This is what appears in the Prefect logs. The names of the users instead of their associated uids are now displayed.

pipeline/flows/matching.py
```diff
@@ -18,25 +18,29 @@
...
-from pipeline.types import EngineId, Match, MatchingInput, User
+from pipeline.types import EngineId, Match, MatchingInput, MatchingOutput, User
 from pipeline.utils.firebase import initialize_firebase_for_prefect
 
 
 @prefect.task
-def compute_matches(inp: MatchingInput, engine_id: EngineId) -> List[Match]:
+def compute_matches(
+    inp: MatchingInput, engine_id: EngineId
+) -> Tuple[MatchingOutput, List[Match]]:
```
vingkan, line 43: Good job updating this!

chunter3:
Given the state of the current matching flow (as seen in the images I've posted above), I cannot really verify the correct behavior of the display function. Therefore, I wrote some tests for it.

vingkan (approved):
Great effort on this pull request! Two things you did really impressed me and there are two things that I think can be improved on:

1. I love that when you ran into a case that you could not verify manually, you wrote a unit test for it - great idea.
2. When faced with a harder version of the problem (sorting the output and handling duplicate names), you used new features of Python and must have looked some things up to debug.

But on the flip side:
1. The display code has become quite complicated, we'll want to simplify it so it's easier to understand, test, and build on in the future.
2. In order to test the code, you created a function that does two different things: either returns metrics or prints code.

I approved this pull request, so go ahead and merge it. For the two pieces of feedback, I'll send you a pull request where I suggest some edits to your code and you will review it. Will give you practice with code review and will let us discuss some ideas on making code more maintainable and testable.

Great work, Chris!

chunter3:
Thank you. Although for the second piece of feedback I believe I know where you're talking about. There's a small print statement I made while testing, but by no means is that meant to be there. I simply forgot to remove it after everything was said and done. But other than that, thank you for your thoughts, Vinesh. I will merge.

Issue #239, opened 2022-07-05
GitPod does not automatically install pre-commit hooks when a previous workspace is re-opened


vingkan:
Probably some funny business with the `init` task not running.

vingkan:
We also found that for some interns, this is not working even when they launch a new workspace from the GitHub repository badge.

I updated the ticket with a suggested action plan.

PR #265, opened 2022-07-12
Fix precommit hook not installing when opening gitpod

notedwin:
Solves #239

### DIFF
--- a/.gitpod.yml
+++ b/.gitpod.yml
@@ -71,6 +71,7 @@ tasks:
   - name: Precommit
     command: |
       source .venv/bin/activate
+      EXPORT PRE_COMMIT_HOME='workspace/.cache/pre-commit'
       pre-commit install
       pre-commit install --install-hooks
   - name: Terminal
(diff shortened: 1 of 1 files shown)


notedwin:
I made changes to spawn a new terminal.

I tried setting the PRE_COMMIT_HOME variable in different places, to accommodate all new terminals having the same pre-commit installation location, however, this did not work :(

I could have tried creating a hard/soft link between the default pre-commit location and somewhere within the workspace.

Pre-commit has locking on the pre-commit directory, which prevents us from spawning a 2nd download task, which makes our short-term solution work well!

.gitpod.yml
```diff
@@ -68,12 +68,15 @@ tasks:
...
+  - name: Precommit
+    command: |
+      source .venv/bin/activate
+      pre-commit install --install-hooks
   - name: Terminal
     init: |
       python3 -m venv .venv
       source .venv/bin/activate
       ./run.sh install-backend
-      pre-commit install
```
vingkan, line 76: Nice! I didn't realize that `pre-commit install --install-hooks` also installs pre-commit. Good to know.

vingkan (approved):
Tested it out, works on my workspace!

Issue #252, opened 2022-07-11
Integrate matching engine analytics into matching flow


vingkan:
## Goal

In #224 and #251, we are adding new metrics to analyze the matching engine. Now we want to calculate those as part of the matching flow and display the results.

For now, we do not need anything too fancy to display the analytics. We will print the results in the Prefect logs of the matching flow.

## Definition of Done

- [ ] Create a dataclass in `pipeline/types` to hold the output of matching engine analytics
- [ ] Write a Prefect task in `pipeline/transform` that computes the metrics by calling each metric function
- [ ] Write a Prefect task in `pipeline/load` that displays the results of the metrics from the analytics output data
- [ ] Integrate the tasks into the matching flow and ensure that the flow runs without errors and displays the correct analytics

## Advice

- In the transform task, it is okay to call each metric function one by one and save their output
- For now, we can skip writing unit tests for the compute task because the individual metrics have unit tests, so adding a new integration tests gives only a small amount of added value
- For now, we can skip writing tests for the display task because the output will be printing to logs, so it is harder to write tests for this than it is to fix bugs that might pop up, also the added value is low because it is displaying information produced by functions that have unit tests
- Adding the `@task` decorator turns a function into a Prefect task, which allows it to be used in a Prefect flow

## Code Pointers

Create a new dataclass for matching engine analytics in this file: https://github.com/scarletstudio/butterfly/blob/e37d49522609e324ee665751a2405aa8dfdb8f49/pipeline/types/engine.py#L45-L47

See an example of a transform task that outputs a dataclass: https://github.com/scarletstudio/butterfly/blob/e37d49522609e324ee665751a2405aa8dfdb8f49/pipeline/transform/matches.py#L16-L19

See an example of a load task that writes information to Prefect logs: https://github.com/scarletstudio/butterfly/blob/e37d49522609e324ee665751a2405aa8dfdb8f49/pipeline/load/validation.py#L13-L30

Add your compute and display tasks to the matching flow after the matching output is ready: https://github.com/scarletstudio/butterfly/blob/e37d49522609e324ee665751a2405aa8dfdb8f49/pipeline/flows/matching.py#L112-L116

Test the matching flow locally by running:

```bash
run flow matching
```

This will not write any results to the database, but it will let you see all the logs and the SUCCESS/FAILURE status.

You can also run the flow in Prefect dashboard and look at the logs tab.

PR #270, opened 2022-07-14
Refactor task and tests to display proposed match counts per user.

vingkan:
Refactor of the display code in #255.

For feedback on how to implement part of #252 in a way that is simpler, more testable, and easier to build on.

### DIFF
--- a/pipeline/load/display_metrics.py
+++ b/pipeline/load/display_metrics.py
@@ -1,3 +1,4 @@
+from operator import itemgetter
 from typing import Dict, List
 
 import prefect
@@ -6,18 +7,17 @@ from prefect import task
 from pipeline.types import MatchingMetrics, MatchingOutput, User, UserId
 
 
-def render_counts_per_user(
-    users: List[User], counts: Dict[UserId, int]
-) -> List[str]:
-    """Helper method to get lines of counts per user in descending order."""
+def render_counts_per_user(users: List[User], counts: Dict[UserId, int]) -> str:
+    """Helper method to get log of counts per user in descending order."""
     display_names = {user.uid: user.displayName for user in users}
     # Sort entries by count in descending order
-    sorted_items = sorted(counts.items(), key=lambda p: p[1], reverse=True)
+    sorted_items = sorted(counts.items(), key=itemgetter(1), reverse=True)
     lines = []
     for uid, count in sorted_items:
         name = display_names.get(uid)
         lines.append(f"{count} - {name} ({uid})")
-    return lines
+    joined = "\n".join(lines)
+    return joined
 
 
 @task
@@ -26,9 +26,7 @@ def display_internal_matching_metrics(
 ):
     """Task to display matching engine metrics."""
     logger = prefect.context.get("logger")
-    lines = render_counts_per_user(
+    proposed_matches_per_user = render_counts_per_user(
         output.users, metrics.n_proposed_matches_per_user
     )
-    logger.info("Matches Proposed per User:")
-    for line in lines:
-        logger.info(line)
+    logger.info(f"\nMatches Proposed per User:\n{proposed_matches_per_user}")
--- a/pipeline/load/display_metrics.py
+++ b/pipeline/load/display_metrics.py
@@ -1,59 +1,34 @@
-import operator
+from typing import Dict, List
 
 import prefect
 from prefect import task
 
-from pipeline.types import MatchingMetrics, MatchingOutput, User
+from pipeline.types import MatchingMetrics, MatchingOutput, User, UserId
+
+
+def render_counts_per_user(
+    users: List[User], counts: Dict[UserId, int]
+) -> List[str]:
+    """Helper method to get lines of counts per user in descending order."""
+    display_names = {user.uid: user.displayName for user in users}
+    # Sort entries by count in descending order
+    sorted_items = sorted(counts.items(), key=lambda p: p[1], reverse=True)
+    lines = []
+    for uid, count in sorted_items:
+        name = display_names.get(uid)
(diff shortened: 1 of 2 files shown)


pipeline/load/display_metrics.py
```diff
@@ -1,59 +1,34 @@
...
+from typing import Dict, List
 
 import prefect
 from prefect import task
 
-from pipeline.types import MatchingMetrics, MatchingOutput, User
+from pipeline.types import MatchingMetrics, MatchingOutput, User, UserId
+
+
+def render_counts_per_user(
```
vingkan, line 9: I created a helper function for three reasons:
1. Easier to test because it returns a list of strings
2. Easier to reuse for future user count metrics because it is not specific to proposed matches per user
3. Easier to read because it separates the names and sorting from the logging code

Think about how to break code down into smaller functions for testing, reuse, and readability. But don't break them down so small that it becomes tedious to jump around and understand what the whole code does.

pipeline/load/display_metrics.py
```diff
@@ -1,59 +1,34 @@
...
+    users: List[User], counts: Dict[UserId, int]
+) -> List[str]:
+    """Helper method to get lines of counts per user in descending order."""
+    display_names = {user.uid: user.displayName for user in users}
+    # Sort entries by count in descending order
+    sorted_items = sorted(counts.items(), key=lambda p: p[1], reverse=True)
```
vingkan, line 15: We don't need to use the `operator.itemgetter` module to sort by counts, we can use a lambda and sort by the second item in the tuple. I would prefer to avoid adding a dependency, but either way, I add a comment to clearly explain what is happening, because the `1` and the `reverse=True` is hard to remember the meaning of.

pipeline/load/display_metrics.py
```diff
@@ -1,59 +1,34 @@
...
 
 import prefect
 from prefect import task
 
-from pipeline.types import MatchingMetrics, MatchingOutput, User
+from pipeline.types import MatchingMetrics, MatchingOutput, User, UserId
+
+
+def render_counts_per_user(
+    users: List[User], counts: Dict[UserId, int]
```
vingkan, line 10: I only pass in `users` and `counts`, since I don't need the rest of the matching output or the metrics output.

pipeline/load/display_metrics.py
```diff
@@ -1,59 +1,34 @@
...
-        + "\n".join(
-            f"{count} - {user}"
-            for user, count in n_proposed_matches_per_user.items()
-        )
+    lines = render_counts_per_user(
+        output.users, metrics.n_proposed_matches_per_user
     )
+    logger.info("Matches Proposed per User:")
+    for line in lines:
+        logger.info(line)
```
vingkan, line 34: Now the display task is very simple: it just logs the lines returned by the helper function.

pipeline/load/display_metrics_test.py
```diff
@@ -1,63 +1,27 @@
...
-        "Liliana": 5,
-        "Marilyn": 4,
-        "Chris (uid: C)": 2,
-        "Ayman": 1,
-    }
-
-    assert list(actual.keys()) == list(expected.keys())
-
-    # Users w/ identical names will be further distinguished with their uid
+    assert actual == expected
```
vingkan, line 27: Now the test is much simpler and easier to read. If your test is becoming very complicated, that is an "instinctive" sign that the function you are testing might be doing too much.

pipeline/load/display_metrics_test.py
```diff
@@ -1,63 +1,27 @@
...
-    actual = display_internal_matching_metrics.run(
-        output, metric, return_metric=True
-    )
-    expected = {
-        "Chris (uid: E)": 6,
-        "Liliana": 5,
-        "Marilyn": 4,
-        "Chris (uid: C)": 2,
-        "Ayman": 1,
-    }
```
vingkan, line 59: Showing the user ID **only** when two users have the exact same name adds unnecessary complexity to the logic. We can distinguish them by always showing the user ID.

When I told you about cases where two users have identical display names, all I wanted you to do was not use display name as a key.

pipeline/load/display_metrics.py
```diff
@@ -1,59 +1,34 @@
...
 
-from pipeline.types import MatchingMetrics, MatchingOutput, User
+from pipeline.types import MatchingMetrics, MatchingOutput, User, UserId
+
+
+def render_counts_per_user(
+    users: List[User], counts: Dict[UserId, int]
+) -> List[str]:
+    """Helper method to get lines of counts per user in descending order."""
+    display_names = {user.uid: user.displayName for user in users}
```
vingkan, line 13: Instead of doing a complicated key replace, we can use a dictionary comprehension to get the name for each user ID, then get it simply by calling:

```python
name = display_names.get(uid)
```

We do not have to worry about cases where the user ID does not have a name, since that would be caused by an error elsewhere in the codebase.

chunter3 (approved):

pipeline/load/display_metrics.py
vingkan: Very cool! I learned something new today. I didn't know [itemgetter is faster](https://stackoverflow.com/a/58171096).

I decided to take the dependency anyway and use itemgetter. It's not a big deal performance-wise, but I like your idea and want to use it as an example.

pipeline/load/display_metrics.py
vingkan: This is a great point! I agree that it would be better to use just one log per metric rather than one log per line.

I revised the display method to return a multi-line string rather than a list of strings, updated the test, and then updated the logging code.

PR #277, opened 2022-07-15
Rare Intents Matching, Pass first test case

nelmajdoub:


pipeline/matching/generators/rare_intents_test.py
```diff
@@ -50,8 +52,175 @@ def test_example():
...
+#             users={"1", "3"},
+#             metadata=MatchMetadata(
+#                 generator="rareIntentsGenerator",
+#                 rareIntents=[expected_intent],
+#             ),
+#         ),
+#     ]
+
+
+# def testFour(): #testing if generator works when given no seekers
```
arushirai1, line 184: Also need to add test cases for when (1) two users can match based on multiple rare intents and (2) when a user has both giving and seeking sides for the same intent, there's not a match generated with this sole user.

arushirai1 (changes requested):

arushirai1 (approved):

PR #283, opened 2022-07-18
Adding more intents to the edit profile page.

mflowers1395:
I added more intents to the edit profile page.


vingkan (approved):
Approved. I left a comment about the story files, you can update that if you want.

PR #287, opened 2022-07-19
Find Match of 3

azuhdz7:


pipeline/utils/helpers.py
```diff
@@ -0,0 +1,10 @@
+from typing import List, Optional
+
+from pipeline.types import Match
+
+
+def find_match_of_3(matches: List[Match]) -> Optional[Match]:
```
vingkan, line 6: Would be good to add a comment that explains what the function does:

```python
def find_match_of_3(matches: List[Match]) -> Optional[Match]:
  """"This function..."""
  for match in matches:
```

pipeline/utils/helpers_test.py
```diff
@@ -0,0 +1,43 @@
+import pytest
+
+from pipeline.types import Match
+from pipeline.utils.helpers import find_match_of_3
+
+
+def test_3_match():
```
vingkan, line 7: Would be good to change the names of the test functions to be more descriptive. For example: `test_returns_match_of_three()`.

vingkan (approved):
Great work @azuhdz7! I left two optional comments about documentation and test naming. You can revise those if you want.

And whenever you are ready, you can merge.

PR #288, opened 2022-07-19
Implemented  similar schedules generator

emmadiamon:


pipeline/matching/generators/similar_schedules.py
```diff
@@ -14,6 +14,48 @@ def __init__(self, min_common: int = 0):
...
+        user_2_i = 1
+
+        while user_2_i < len(inp.users):
+            user_1 = inp.users[user_1_i]
+            user_2 = inp.users[user_2_i]
+            for availability1 in user_1.schedule:
+                for availability2 in user_2.schedule:
+                    if availability1 == availability2:
+                        # if we have already saved this match, update the metadata
+                        if (user_1.uid, user_2.uid) in potential_matches:
```
vingkan, line 29: Repeating `(user_1.uid, user_2.uid)` each time you need to check if the pairing exists is making this code cluttered and harder to read. One way to improve this is to save the key as a variable, then use the variable.

For example:

```python
user_pair = (user_1.uid, user_2.uid)
...
if user_pair in potential_matches:
...
match = potential_matches[user_pair]
...
```

Try that out!

pipeline/matching/generators/similar_schedules.py
```diff
@@ -14,6 +14,48 @@ def __init__(self, min_common: int = 0):
...
+                                    generator="similarSchedulesGenerator",
+                                    score=1,
+                                    matchingAvailability=[availability1],
+                                ),
+                            )
+            if user_2_i == len(inp.users):
+                user_1_i += 1
+                user_2_i = user_1_i + 1
+            else:
+                user_2_i += 1
```
vingkan, line 52: There could be a simpler way to loop over all the users and their matching availabilities, but I like what you did here because it seems to be working. Let's make sure it works by adding more test cases that will cause both of these code blocks to run.

In the future, if we are sure the code is correct and has tests, then we could try some ways to make the looping simpler.

vingkan (changes requested):
This is a great start, @emmadiamon! I'm glad we were able to push what you had so far, fix the mypy warnings, and update your branch with the latest code from main.

I have requested two changes, please check out my comments and give those a try. Message me on Slack if you have any questions or run into issues.

pipeline/matching/generators/similar_schedules.py
vingkan: Good work!

pipeline/matching/generators/similar_schedules_test.py
```diff
@@ -42,9 +42,8 @@ def test_example():
             ),
         ),
     ]
-    # TODO: Uncomment this assertion and delete the empty list assert
-    # assert actual == expected
-    assert actual == []
+
+    assert actual == expected
```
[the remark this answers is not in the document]
vingkan: We decided that we need these three tests:
1. users with similar schedules (already set up)
2. users with no similar schedules
3. increase min_common

And in some of those tests, you should add availability that is not similar, to make sure that only common availability is included in the output. For example:

- if user A is free on Monday and Tuesday
- and user B is free on Tuesday and Wednesday
- then the match should only have Tuesday

vingkan (approved):
Great work, Emma! We can merge this now and then you can add the last test case later on.

PR #290, opened 2022-07-19
Added notification dataclass and the logic for various notifications

chunter3:
Not all notification logic has been written. WIP


pipeline/types/notifications.py
```diff
@@ -0,0 +1,12 @@
...
+from pipeline.types.chat import ChatData
+from pipeline.types.match import Match
+
+
+@dataclass
+class Notification:
+    # a single match of n-users (where n >= 2)
+    match: Match
+    # relevant chat data between the users in the given match
+    chatdata: ChatData
```
vingkan, line 12: `ChatData` should contain the same information as `Match`, so I don't think you need both.

pipeline/notifications/check_message.py
```diff
@@ -0,0 +1,25 @@
...
+
+class Notifier:
+    def __init__(self):
+        pass
+
+    def check_msg_notifier(self, notif: Notification) -> List[User]:
+
+        # check if all users in match have a message
+        # if there are users who don't have a message, return a list of them for notification
+        users_with_msg = []
```
vingkan, line 17: Use a set here, since there is no need to store a user multiple times.

Otherwise, if there are a thousand messages in the chat, you might end up with a thousand values in `users_with_msg`.

Important to think about when to use list vs set vs dict vs tuple.

pipeline/notifications/check_message.py
```diff
@@ -0,0 +1,25 @@
...
+        users_to_notify = []
+        for msg in notif.chatdata.messages:
+            if msg.from_user in notif.match.users:
```
vingkan, line 20: Good thinking. I would have assumed that only users who are in the match could have sent a message to the chat, but this is a good check.

When you refactor to use only `ChatData` instead of both `ChatData` and `Match`, you can check `chatdata.participants` for the user ID.

pipeline/notifications/check_message.py
```diff
@@ -0,0 +1,25 @@
+from typing import List
+
+import prefect
+from prefect import task
+
+from pipeline.types import Notification, User
+
+
+class Notifier:
```
vingkan, line 9: If you are going to have a common class that specific implementations inherit from, then save the general name `Notifier` for the parent class, and have each implementation use a specific name, like `NewMessageNotifier`.

pipeline/notifications/check_message.py
```diff
@@ -0,0 +1,25 @@
...
+    def __init__(self):
+        pass
+
+    def check_msg_notifier(self, notif: Notification) -> List[User]:
```
vingkan, line 13: I think you can just pass in the chat data here.

pipeline/types/notifications.py
```diff
@@ -0,0 +1,12 @@
+from dataclasses import dataclass
+
+from pipeline.types.chat import ChatData
+from pipeline.types.match import Match
+
```
vingkan, line 5: Consider adding a type for the output notification. What information do we need to give the system that is sending the notification?

- User to send to
- Email of user to send to (this is not on the `User` dataclass currently, but we will add it shortly)
- Content of notification
- Etc...

pipeline/notifications/check_message_test.py
```diff
@@ -0,0 +1,56 @@
+from datetime import datetime
+
+import pytest
+
+from pipeline.notifications import Notifier
+from pipeline.types import ChatData, Match, Message, Notification, User
+
```
vingkan, line 7: Good tests! Would recommend adding a case where both users have sent messages, so there are no users to notify.

pipeline/notifications/check_message.py
vingkan, line 13: We may want to rename this method to better describe what it does. Since it returns users, we could call it `get_users_to_notify()`.

You might want to add another method (or combine into this method) something like `get_notification()` which returns all the information needed to actually send the notification, like email, subject, etc..

What information will another engineer need to write a new notification implementation for one of the other use cases we want to support? Will these method names still be relevant to them? When designing a parent and child classes, these are topics to think about, so that the parent class helps clarify the logic of the child classes (implementations), rather than forcing the child classes to conform to a structure that does not make sense for them.

pipeline/notifications/check_message.py
```diff
@@ -0,0 +1,25 @@
...
+
+        # check if all users in match have a message
+        # if there are users who don't have a message, return a list of them for notification
+        users_with_msg = []
+        users_to_notify = []
+        for msg in notif.chatdata.messages:
```
vingkan, line 19: One big thing to think about (we should save this for a future PR): how will you deal with the timing of the pipeline runs?

For example, suppose user A messages the chat with user B and user B does not respond for the whole week. If we run our notification pipeline every night, we might send user B the same notification every day for the whole week.

Here are some ideas:
- A. Filter to only consider messages from the last 24 hours
- B. Save the timestamp of the last pipeline run and filter to only consider messages since then
- C. Store previously sent notifications and check against them to avoid creating a duplicate notification (or a notification of the same type to the same user)

Can you think of other approaches? What are the pros/cons of these approaches? Which would you choose to implement?

vingkan (commented):
Good start! After you make some edits, we can merge this so that you have code checked in as a starting point.

Some of the comments I left are edits requested to merge this PR, others are comments for feedback on the project as a whole to be addressed in future PRs. Let me know if you are not sure which is which.

pipeline/notifications/check_message.py
chunter3: I had a little trouble understanding what you meant by this comment, but now I understand. It be more or less similar to how it is in pipeline/matching/generators. And I agree that be for the best going forward for future implementations.

pipeline/types/notifications.py
chunter3: This shouldn't be necessary anymore, right? Because the User datatype has been refactored to include email and the like. We'd just return the User object and, if that information is needed, we pull it directly from it.

pipeline/notifications/check_message.py
vingkan: What do you think about this @chunter3 ?

vingkan (changes requested):
Looking good! I requested a few small changes, then we can merge this.

I want you to merge this PR sooner because other interns are ready to start building the notifications for the other use cases. We should merge the most important things (like creating the notifications folder, creating the notifications types, and merging a simple even if imperfect example) sooner so that others are not blocked.

This is a good practice for software development: break your work into smaller chunks and merge those as small PRs, rather than trying to do a huge amount of work in one pull request. More details on this advice in Google's engineering practices on [Small CLs ](https://google.github.io/eng-practices/review/developer/small-cls.html) (CL is what Google calls a PR).

pipeline/notifications/check_message.py
vingkan: Awesome! Great analysis. I agree on choosing option B.

If it's easier to build, we can start by implementing option A, with a variable number of hours, that way even if we run the pipeline weekly, we can adjust it.

Then, if we have time, we can refactor the pipeline to use option B.

vingkan (approved):
Great work! This is ready to merge.

vingkan:
This is the first PR for project #303.

PR #296, opened 2022-07-20
added google analytics

phodonou:


frontend/src/app/data/User.js
```diff
@@ -19,6 +21,7 @@ export async function fetchUserData(userId) {
  * Retrieves data directly from Firebase, requires user to be authenticated.
  */
 export function getUserData(userId) {
+    getAnalytics()
```
vingkan, line 24: Don't import and call it in this file because this will only run if this function is called. Instead, you can import and call it in App.jsx.

vingkan (changes requested):
Left one comment!

vingkan (approved):
Approving it so you can merge after making that one change. Let me know if it works in prod! You can check GitHub or the #z-alerts channel on Slack to see when the GitHub pages deploy starts and finish.

PR #298, opened 2022-07-20
Add all user fields from database to User dataclass

vingkan:
This will unblock work on the notifications pipeline by giving us the user's email.

- [x] Add remaining database fields to `User` dataclass
  - Email, photo URL, latest login, and community memberships
- [x] Update extract code to rename columns from camel case to snake case
- [x] Update extract unit tests so that the `is_active` and `communities` columns are no longer dropped
- [x] Update extract unit tests to test that specified columns are renamed
- [x] Update extract unit tests to test that all columns from database are retained in DataFrame
- [x] Update extract unit tests to use `np.nan` for missing values
- [x] Update extract unit tests to use `RawCommunityMembership`
- [x] Move raw community membership types to be with other raw types
- [x] Update transform code to save community memberships as a dataclass
- [x] Update transform unit tests to verify community memberships
- [x] Test by running the matching flow for all communities

### DIFF
--- a/pipeline/extract/users.py
+++ b/pipeline/extract/users.py
@@ -4,6 +4,11 @@ from prefect import task
 
 from pipeline.types import Community, UserCommunityMemberships
 
+USER_COLUMN_RENAMES = {
+    "latestLogin": "latest_login",
+    "photoURL": "photo_url",
+}
+
 
 def is_active_in_community(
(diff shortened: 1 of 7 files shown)


chunter3 (approved):
Looks good. I see no issues with this.

PR #301, opened 2022-07-21
limited schedules

eeerie:


pipeline/matching/generators/limited_schedules.py
```diff
@@ -7,13 +7,78 @@
...
+            for user in inp.users:
+                if (
+                    limited_user.uid != user.uid
+                ):  # check that we are not looking at the current limited_user
+                    # compare limited user schedule to schedule of user in all users
+                    for l_availability in limited_user.schedule:
+                        for availability in user.schedule:
+                            if (
+                                l_availability == availability
+                            ):  # if the availabilities in limited user schedule and other user schedule are the same, then we have a match!
```
vingkan, line 42: Right now, I think what you are doing is getting just one time the two users have in common. We want to get all the times they have in common. It should be a small list since one user has limited availability. But, since the parameter `max_available` can change, we shouldn't assume that it will always be `1`.

I would recommend creating a dictionary like `common_availability` where the key is the set of the two user IDs and the value is a list of the availabilities they have in common. When you go through this loop, you can append the matching availabilities to that list.

pipeline/matching/generators/limited_schedules.py
```diff
@@ -7,13 +7,78 @@
...
+                                    user.uid,
+                                ) in matches:  # add availability to availability metadata
+                                    pass
+                                else:
+                                    # add first match of this set of users to dictionary
+                                    # TODO: add the availability to this first match
+                                    matches[
+                                        {limited_user.uid, user.uid}
+                                    ] = Match(
+                                        users={limited_user.uid, user.uid}
```
vingkan, line 54: When you are ready to create the matches, you can add the availabilities the two users have in common like this:

```python
# Put this at the top of your file to import the metadata (or add it to the list of types already being imported)
from pipeline.types import MatchMetadata

Match(
  users={uid1, uid2},
  metadata=MatchMetadata(
    generator=GENERATOR_LIMITED_SCHEDULES,
    availability=availability
  )
)
```

The variable `availability` should be a list of the matching availabilities between the users, from their schedules.

vingkan (changes requested):
Glad you could open this pull request, @eeerie!

I left comments on next steps, let me know if you have any questions. Hopefully you can work on some of these before our next meeting.

pipeline/matching/generators/limited_schedules.py
```diff
@@ -7,13 +7,78 @@
...
+    def __init__(self, max_available: int = 1):
         # Maximum number of available slots in a user's schedule for them to be
         # eligible for a limited availability match.
         self.max_available = max_available
         super().__init__(name=GENERATOR_LIMITED_SCHEDULES)
 
     def generate(self, inp: MatchingInput) -> Iterator[Match]:
         # TODO: Implement your generator
-        inp.logger.info("Luke Arrington was here!")
+
```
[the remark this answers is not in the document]
vingkan: Great work! We also improved the indentation issues by using a `continue` in the loop.

vingkan (approved):
Great work, Luke! We confirmed that the tests pass and the pipeline passes, so we can merge this.

You can add more test cases in a future pull request.

pipeline/matching/generators/limited_schedules.py
```diff
@@ -1,19 +1,56 @@
...
+        limited_availability = []
+        for user in inp.users:
+            if len(user.schedule) <= self.max_available:
+                limited_availability.append(user)
+
+        users_availability: Dict[Tuple, List] = {}
+        # compare limited users to all users
+        for limited_user in limited_availability:
+            for user in inp.users:
+                userid_tuple = (limited_user.uid, user.uid)
```
vingkan, line 38: There's also a subtle issue here that if both users are limited users, we might create two versions of the same match, one where the first user comes first in the tuple, and one with the opposite order.

This is okay because the matching engine ignores duplicate matches. But we can fix this in the future when you write more test cases.

Issue #281, opened 2022-07-18
Variety ranker encountered an exception during the matching flow run


vingkan:
To reproduce:
- Run the matching flow from Prefect dashboard with:
  - Community: sds
  - engine: main
  - Force: false
  - Release: 2022-07-18

This error does not arise on the demo community, but it does for the sds community.

Error log:

```
Task 'compute_matches': Exception encountered during task execution!
Traceback (most recent call last):
  File "/workspace/butterfly/.venv/lib/python3.8/site-packages/prefect/engine/task_runner.py", line 880, in get_task_run_state
    value = prefect.utilities.executors.run_task_with_timeout(
  File "/workspace/butterfly/.venv/lib/python3.8/site-packages/prefect/utilities/executors.py", line 468, in run_task_with_timeout
    return task.run(*args, **kwargs)  # type: ignore
  File "/workspace/butterfly/./pipeline/flows/matching.py", line 53, in compute_matches
    output = engine.run(inp)
  File "/workspace/butterfly/pipeline/matching/core/engine.py", line 42, in run
    ranked_copy: List[Match] = list(ranked)
  File "/workspace/butterfly/pipeline/matching/rankers/quantity.py", line 17, in rank
    yield from proposed
  File "/workspace/butterfly/pipeline/matching/rankers/variety.py", line 30, in rank
    yield from sorted(
  File "/workspace/butterfly/pipeline/matching/rankers/variety.py", line 32, in <lambda>
    key=lambda match: dup_gencheck(inp.recent_matches, match),
  File "/workspace/butterfly/pipeline/matching/rankers/variety.py", line 12, in dup_gencheck
    match.metadata.generator == recent.metadata.generator
AttributeError: 'dict' object has no attribute 'generator'
```

PR #302, opened 2022-07-21
Fixed matching flow exception

dfg-13:
#281 

The initial bug was that MatchMetadata was running into errors because there was no "get" function. This prompted a change in the variety.py file, more specifically, ```match.metadata.generator == recent.metadata.generator```. I later found it was an issue in the way matches were extracted. All the changes that were made worked, but mypy prevented me from committing.

### DIFF
--- a/pipeline/transform/matches.py
+++ b/pipeline/transform/matches.py
@@ -30,7 +30,6 @@ def convert_matches_from_df(df: pd.DataFrame) -> List[Match]:
     cols = set(df.columns)
     field_names = [f.name for f in dataclasses.fields(Match) if f.name in cols]
     match_dicts = df[field_names].to_dict(orient="records")
-    # matches = [Match(**u) for u in match_dicts]
     matches = []
     for record in match_dicts:
         raw_metadata = record.get("metadata", {})
--- a/pipeline/transform/matches.py
+++ b/pipeline/transform/matches.py
@@ -6,7 +6,13 @@ import prefect
 from prefect import task
 
 from pipeline.matching.utils import generate_keys
-from pipeline.types import Community, Match, MatchingInput, ReleaseTag
+from pipeline.types import (
+    Community,
+    Match,
+    MatchingInput,
+    MatchMetadata,
+    ReleaseTag,
+)
 
 
 class MissingReleaseTagException(Exception):
@@ -24,7 +30,10 @@ def convert_matches_from_df(df: pd.DataFrame) -> List[Match]:
     cols = set(df.columns)
     field_names = [f.name for f in dataclasses.fields(Match) if f.name in cols]
     user_dicts = df[field_names].to_dict(orient="records")
-    return [Match(**u) for u in user_dicts]
+    matches = [Match(**u) for u in user_dicts]
+    for match in matches:
+        match.metadata = MatchMetadata(str(match.metadata))
+    return matches
 
 
 @task
--- a/pipeline/transform/matches.py
+++ b/pipeline/transform/matches.py
@@ -29,10 +29,15 @@ def get_matching_input(**kwargs) -> MatchingInput:
 def convert_matches_from_df(df: pd.DataFrame) -> List[Match]:
     cols = set(df.columns)
     field_names = [f.name for f in dataclasses.fields(Match) if f.name in cols]
-    user_dicts = df[field_names].to_dict(orient="records")
-    matches = [Match(**u) for u in user_dicts]
-    for match in matches:
-        match.metadata = MatchMetadata(str(match.metadata))
+    match_dicts = df[field_names].to_dict(orient="records")
+    # matches = [Match(**u) for u in match_dicts]
+    matches = []
+    for record in match_dicts:
+        raw_metadata = record.get("metadata", {})
+        parsed_metadata = MatchMetadata(**raw_metadata)
+        record.pop("metadata", None)
+        match = Match(**record, metadata=parsed_metadata)
+        matches.append(match)
     return matches
 
 
--- a/pipeline/transform/matches_test.py
+++ b/pipeline/transform/matches_test.py
@@ -4,13 +4,15 @@ from pipeline.transform.matches import (
     convert_matches_from_df,
     filter_recent_matches,
 )
-from pipeline.types import Match
+from pipeline.types import Match, MatchMetadata
 
 
 def test_convert_matches_from_df():
     df = pd.DataFrame([{"users": {"A", "B"}, "extraData": "ignore"}])
     actual = convert_matches_from_df.run(df)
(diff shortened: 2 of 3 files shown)


pipeline/transform/matches.py
```diff
@@ -24,7 +30,10 @@ def convert_matches_from_df(df: pd.DataFrame) -> List[Match]:
     cols = set(df.columns)
     field_names = [f.name for f in dataclasses.fields(Match) if f.name in cols]
     user_dicts = df[field_names].to_dict(orient="records")
```
vingkan, line 32: This variable can be renamed to `match_dicts`, it was my mistake that it was called `user_dicts` because I just copied the code for the user transform :)

pipeline/transform/matches.py
```diff
@@ -24,7 +30,10 @@ def convert_matches_from_df(df: pd.DataFrame) -> List[Match]:
...
-    return [Match(**u) for u in user_dicts]
+    matches = [Match(**u) for u in user_dicts]
+    for match in matches:
+        match.metadata = MatchMetadata(str(match.metadata))
```
vingkan, line 35: Why are we casting `match.metadata` to a string here? This is also related to why the unit tests for this pull request are failing.

pipeline/transform/matches.py
```diff
@@ -24,7 +30,10 @@ def convert_matches_from_df(df: pd.DataFrame) -> List[Match]:
     cols = set(df.columns)
     field_names = [f.name for f in dataclasses.fields(Match) if f.name in cols]
     user_dicts = df[field_names].to_dict(orient="records")
-    return [Match(**u) for u in user_dicts]
+    matches = [Match(**u) for u in user_dicts]
```
vingkan, line 33: It would be better to parse the metadata before the `Match` objects are created, rather than editing it after the fact. You could change this list comprehension to a loop:

```python
matches = []
for record in match_dicts:
  # Some matches don't have metadata, so we should get an empty dictionary in that case
  raw_metadata = record.get("metadata", {})
  # There may be additional parsing needed for some of the more complicated types, but this should work for now, since you only depend on the generator field
  parsed_metadata = MatchMetadata(**raw_metadata)
  # We can override the metadata field with our new parsed value like this:
  match = Match(**record, metadata=parsed_metadata)
return matches
```

vingkan (changes requested):
Thanks for taking on this bug! Before the interns came, we didn't have any metadata in any production matches, so I completely missed this issue.

I left some comments, also please update the unit tests to add a case where metadata should be included in the transform. If the unit test takes too long, just ask me to approve the logic and I can add the test later.

pipeline/transform/matches.py
dfg-13: I tried different combinations of types and this was the only thing that worked. In hindsight, this was probably not the right thing to do. Though it is fixed now. :)

pipeline/transform/matches.py
dfg-13: I tried this code and ran into a problem regarding match having two metadata's. I found that deleting the metadata from the record after parsing worked along with appending the match to the matches list.

pipeline/transform/matches_test.py
```diff
@@ -4,13 +4,15 @@
...
+from pipeline.types import Match, MatchMetadata
 
 
 def test_convert_matches_from_df():
     df = pd.DataFrame([{"users": {"A", "B"}, "extraData": "ignore"}])
     actual = convert_matches_from_df.run(df)
-    expected = [Match(users={"A", "B"})]
+    expected = [
+        Match(users={"A", "B"}, metadata=MatchMetadata(generator="blank"))
+    ]
```
notedwin, line 15: If you remove what we added, the test fails because the match metadata's generator field equals something related to magic mock.
I did not fully understand the problem.

Screenshot 2022-07-23 at 12 26 55 PM

notedwin (approved):

pipeline/transform/matches_test.py
vingkan: Yeah, this is a weird error. There's no mock in this test at all. At least it's working now :)

vingkan, line 15: Can you add another test case? This only tests that it works when there is no metadata passed in (which is an important case).

We also need a case where there is metadata in the input dataframe to transform.

pipeline/transform/matches.py
```diff
@@ -23,8 +29,16 @@ def get_matching_input(**kwargs) -> MatchingInput:
...
     field_names = [f.name for f in dataclasses.fields(Match) if f.name in cols]
-    user_dicts = df[field_names].to_dict(orient="records")
-    return [Match(**u) for u in user_dicts]
+    match_dicts = df[field_names].to_dict(orient="records")
+    # matches = [Match(**u) for u in match_dicts]
+    matches = []
+    for record in match_dicts:
+        raw_metadata = record.get("metadata", {})
+        parsed_metadata = MatchMetadata(**raw_metadata)
+        record.pop("metadata", None)
```
vingkan, line 38: Great! Glad you figured this out.

vingkan (approved):
Thanks for fixing this! I confirmed that the pipeline works. We can merge this ASAP to fix the pipeline.

I requested one more test case to be added, you can add it, or you can leave it for me to do, or to be done in a future PR. Anything works!

pipeline/transform/matches_test.py
```diff
@@ -4,13 +4,31 @@
...
+
+
+def test_convert_matches_from_df_2():  # metadata is passed using the blue generator
+    df = pd.DataFrame([{"users": {"A", "B"}, "extraData": "ignore"}])
+    actual = convert_matches_from_df.run(df)
+    actual[
+        0
+    ].metadata.generator = (
+        "blueGenerator"  # used to change the generator from 'blank'
+    )
```
vingkan, line 26: You should add the metadata in the input, like this:

```python
match_data = [{"users": {"A", "B"}, "metadata": {"generator": "blueGenerator"}}]
df = pd.DataFrame(match_data)
```

Right now, you are editing the actual value after the code runs, so of course it will pass, because you you changed it to the right value ;) but then we are not actually testing that the code works as expected.

Also, give the test a more descriptive name:

```python
def test_convert_matches_from_df_with_metadata():
  # metadata is passed using the blue generator
```

vingkan (changes requested):

pipeline/transform/matches_test.py
dfg-13: This makes a lot more sense than what I initially had :). My initial thinking was that `actual` was the issue hence why I went and edited the value in order for it to pass the tests.

vingkan (approved):
Excellent! Great work fixing this, David!

PR #304, opened 2022-07-21
Implementation of Similar_Interests

bridgettorres:


pipeline/matching/generators/similar_interests.py
```diff
@@ -1,18 +1,50 @@
...
-        yield from []
+        listUsers = inp.users
+        if listUsers == []:
+            matchUsers = {"Empty", "Empty"}
+            metadata = MatchMetadata(
+                generator=GENERATOR_SIMILAR_INTERESTS,
+                score=0.0,
+                commonInterests=[],
+            )
+            yield Match(users=matchUsers, metadata=metadata)
```
vingkan, line 24: You don't have to yield an empty match if there are no users. In fact, this will cause an error in the matching engine because `"Empty"` will not be a valid user ID.

In this case, you can either do:

```python
if listUsers == []:
  return iter([])
```

```python
if listUsers == []:
  return
  yield
```

More discussion on [this StackOverflow post](https://stackoverflow.com/a/6266586).

pipeline/matching/generators/similar_interests.py
```diff
@@ -1,18 +1,50 @@
...
+        for user_one, user_two in itertools.combinations(listUsers, r=2):
```
vingkan, line 25: Good idea to use itertools! It makes the code have less deep indentation than if you used nested loops.

pipeline/matching/generators/similar_interests.py
```diff
@@ -1,18 +1,50 @@
...
+            firstInt = user_one.interests
+            secondInt = user_two.interests
+            list_of_interest_user_one = [
+                indivInterest.name.lower() for indivInterest in firstInt
```
vingkan, line 29: We should use the interest `code`, not `name` to match, since the names may change over time, but the codes do not.

And there is no need to lowercase the codes, but it was a good instinct to do data cleaning.

pipeline/matching/generators/similar_interests.py
```diff
@@ -1,18 +1,50 @@
...
+            interest_two = set(list_of_interest_user_two)
+            unique_interest = (interest_one).union(interest_two)
+            common_interest = (interest_one).intersection(interest_two)
+            calculate_score = (
+                (len(common_interest) / len(unique_interest))
+                if (len(unique_interest)) != 0
+                else 0.0
+            )
+            if len(common_interest) < self.min_common:
+                calculate_score = 0.0
```
vingkan, line 44: If there are fewer than `min_common` interests, we should not yield a match at all.

pipeline/matching/generators/similar_interests.py
```diff
@@ -1,18 +1,50 @@
...
+            ]
+            interest_one = set(list_of_interest_user_one)
+            interest_two = set(list_of_interest_user_two)
+            unique_interest = (interest_one).union(interest_two)
+            common_interest = (interest_one).intersection(interest_two)
+            calculate_score = (
+                (len(common_interest) / len(unique_interest))
+                if (len(unique_interest)) != 0
+                else 0.0
+            )
```
vingkan, line 42: Nice scoring! You could move this into its own function if you want.

pipeline/matching/generators/similar_interests_test.py
```diff
@@ -40,9 +41,163 @@ def test_example():
...
+    )
+    generator = SimilarInterestsGenerator(min_common=1)
+    actual = list(generator.generate(inp))
+
+    expected = [
+        Match(
+            users={"1", "2"},
+            metadata=MatchMetadata(
+                generator="similarInterestsGenerator",
+                score=(1 / 5),
```
vingkan, line 111: Nice test case! Good idea to keep this as a fraction instead of the raw float, makes testing more reliable.

pipeline/matching/generators/similar_interests_test.py
```diff
@@ -40,9 +41,163 @@ def test_example():
...
+        Match(
+            users={"1", "2"},
+            metadata=MatchMetadata(
+                generator="similarInterestsGenerator",
+                score=(0.0),
+                commonInterests=[],
+            ),
+        ),
+    ]
+    assert actual == expected
```
vingkan, line 203: Great test cases! Very thorough.

vingkan (changes requested):
Great work @bridgettorres!

I requested some changes before this can be merged. Let me know if you have any questions.

pipeline/matching/generators/similar_interests.py
```diff
@@ -1,18 +1,50 @@
...
+            if len(common_interest) < self.min_common:
+                calculate_score = 0.0
+            metadata = MatchMetadata(
+                generator=GENERATOR_SIMILAR_INTERESTS,
+                score=calculate_score,
+                commonInterests=list(sorted(common_interest)),
+            )
+            yield Match(users={user_one.uid, user_two.uid}, metadata=metadata)
```
BoominHutJr, line 50: Code is very easy to read and understand, LGTM.

pipeline/matching/generators/similar_interests_test.py
```diff
@@ -40,9 +41,163 @@ def test_example():
...
+        ],
+        recent_matches=[],
+        interests=[hiking, swimming, painting, reading],
+    )
+    generator = SimilarInterestsGenerator(min_common=1)
+    actual = list(generator.generate(inp))
+
+    expected = [
+        Match(
+            users={"1", "2"},
```
BoominHutJr, line 69: if there are no common interests, should a match still be generated?

pipeline/matching/generators/similar_interests_test.py
```diff
@@ -40,9 +41,127 @@ def test_example():
...
+            ),
+            User(uid="2", displayName="B", interests=[painting]),
+        ],
+        recent_matches=[],
+        interests=[hiking, swimming, painting, crocheting, boxing],
+    )
+    generator = SimilarInterestsGenerator(min_common=1)
+    actual = list(generator.generate(inp))
+
+    expected = [
```
erikqu, line 97: I like the consistency here.

erikqu (approved):

pipeline/matching/generators/similar_interests.py
```diff
@@ -1,18 +1,44 @@
...
+            interest_two = set(list_of_interest_user_two)
+            unique_interest = (interest_one).union(interest_two)
+            common_interest = (interest_one).intersection(interest_two)
+            calculate_score = (
+                (len(common_interest) / len(unique_interest))
+                if (len(unique_interest)) != 0
+                else 0.0
+            )
+            if len(common_interest) < self.min_common:
+                return iter([])
```
vingkan, line 38: There is an important difference between this case and the case above.

In the first case, we have no users to match, so we should not yield any matches.

In this case, **only** this pair of users have no common interests, so we should skip only their match and keep going on to consider other matches.

The code you have here currently will return an empty iterator and stop considering more matches. Instead, you can use the loop keyword `continue`, which means to skip the rest of the code in this iteration and move on to the next iteration.

vingkan (changes requested):
Great work! I requested one fix, then you can merge this.

vingkan (approved):
LGTM!

PR #322, opened 2022-07-23
Updated NotificationInfo dataclass; refactored get_users_to_notify fu…

chunter3:
…nction; added Notifier parent class


pipeline/types/notification.py
```diff
@@ -8,6 +9,10 @@
 class NotificationInput:
     # relevant chat data between the users in the given match
     chatdata: ChatData
+    # datetime to inform Notifier class when pipeline is being run
+    run_datetime: datetime.datetime
+    # the type of notification
+    notification_type: str = ""
```
vingkan, line 15: Let's make `notification_type` an enum instead, that way we can define explicit types and see all the options. Search the codebase for `Enum` to see some examples.

pipeline/types/notification.py
```diff
@@ -8,6 +9,10 @@
...
 
 
 @dataclass
```
vingkan, line 18: Since we are splitting up the data and presentation of Notifications, you can remove `title` and `content` from the `NotificationInfo` class.

Instead, you can add any custom metadata fields you want for the new message reminder. Like the names of the users who messaged you.

pipeline/notifications/check_message_test.py
```diff
@@ -34,23 +44,26 @@ def test_more_users_in_match():
...
+        chatdata=test_chatdata, run_datetime=datetime.now()
+    )
     check_notifier = NewMessageNotifier()
-    actual = check_notifier.get_users_to_notify(test_chatdata)
+    actual = check_notifier.get_users_to_notify(test_notif_inp)
     assert actual == [
-        User(uid="A", displayName="Ayman"),
-        User(uid="D", displayName="Dinora"),
-        User(uid="E", displayName="Erik"),
+        NotificationInfo(name="Ayman", email="ayman@iit.edu"),
```
vingkan, line 64: I just realized that NotificationInfo doesn't have user ID, which we will need for simple filtering later on.

How about just having a `User` class nested in NotificationInfo?

```
@dataclass
class NotificationInfo:
  recipient: User
```

Could be called `recipient` or `for_user` or just `user`, though the last one might be too vague.

pipeline/matching/core/notifier.py
```diff
@@ -0,0 +1,13 @@
+from typing import List
+
+from pipeline.types import NotificationInfo, NotificationInput
+
+
+class Notifier:
```
vingkan, line 6: Can you add some documentation for the class and the method?

```
class Notifier:
    """"
    Write a description of the parent class here.
    
    And any notes that other classes that inherit from this class should know about.
    """

    def __init__(self, name: str):
        self.name = name

    def get_notifications(
        self, inp: NotificationInput
    ) -> List[NotificationInfo]:
        """"
        Write a description of this method here.
        """"
        pass
```

Explain that:
- A notifier only works on one chat at a time
- The return type of `get_notifications()` is a list because there could be more than one notification generated per chat, it could also be empty

pipeline/notifications/check_message.py
```diff
@@ -3,23 +3,30 @@
 import prefect
 from prefect import task
 
-from pipeline.types import ChatData, NotificationInfo, User
+from pipeline.types import NotificationInfo, NotificationInput
 
 
 class NewMessageNotifier:
```
vingkan, line 9: You can inherit the parent class like this:

```python
class NewMessageNotifier(Notifier):
```

Don't forget to import `Notifier`.

vingkan (changes requested):
Great progress, Chris! This will set up other developers to start writing notifiers.

I requested some changes so that this change can give them a good foundation.

pipeline/notifications/check_message_test.py
chunter3: Yeah I feel we can eliminate the "name" and "email" fields in favor of just having a "recipient" field that takes a User object. I still believe we should keep the "title" and "content" fields unless there's a reason those are superfluous as well.

pipeline/types/notification.py
chunter3: That sounds good. Although besides names of the users who messsaged, I really can't think of any other relevant metadata. If you, or even the other interns, have any other ideas for fields I'll gladly add them refactor the code accordingly

pipeline/notifications/check_message.py
```diff
@@ -3,23 +3,30 @@
...
                 users_with_msg.add(msg.from_user)
-        for uid, user in chatdata.participants.items():
+        for uid, user in inp.chatdata.participants.items():
             if uid not in users_with_msg:
-                users_to_notify.append(user)
+                notif_info_obj = NotificationInfo(
+                    name=user.displayName,
+                    email=user.email,
+                    title=inp.chatdata.title,
+                )
```
chunter3, line 30: I **would** use Counter here since it's O(n) compared to sorted's O(nlog(n)). But seeing as it's highly unlikely that n would be large (n being the number of users in a given match, which realistically should be, at most, 4), I believe it's fine to forgo some efficency. Plus we won't have to include a dependency (i.e. "from collections import Counter")

pipeline/types/notification.py
```diff
@@ -1,22 +1,34 @@
...
 
 from pipeline.types.chat import ChatData
 from pipeline.types.match import Match
+from pipeline.types.user import User
+
+
+class NotificationType(Enum):
+    # An enum representation of all the notification types devised so far; subject to change if more types are added
+    CHECK_MATCH = "New_match"
+    CHECK_MSG = "New_message"
```
vingkan, line 14: Spell out the whole word MESSAGE here, no need to abbreviate.

Also, can you make the enum strings match the name of the enum value, but lowercase? So CHECK_MESSAGE would be "check_message"

pipeline/types/notification.py
vingkan: This looks good! Except, notification_type should be a field of NotificationInfo, not NotificationInput.

And as other fields are needed here, we can add them!

pipeline/notifications/check_message.py
vingkan: Counter wouldn't be necessary because there's nothing to count. That's why you use a set or a dictionary.

Do you need to return a list here? Can messengers be a set? The main reason to convert a set to a list is if order matters. For testing, order doesn't matter when comparing two sets. Maybe you need order when you get to writing the email for the notification, in that case, you could let the display layer handle ordering. It's not a big deal since group chats are 2-3 users and you'll only get this notification if someone didn't message the chat at all, so we're talking about a messagers having only 1-2 values.

pipeline/matching/core/notifier.py
```diff
@@ -0,0 +1,27 @@
+from typing import List
```
vingkan, line 1: I just noticed this is in matching, it should go in pipeline/notifications.

vingkan (changes requested):
Almost there! Requested some minor fixes and then will re-review to approve.

pipeline/notifications/check_message.py
```diff
@@ -3,23 +3,36 @@
...
+        names_of_users_with_msg = set()
         users_to_notify = []
-        for msg in chatdata.messages:
-            if msg.from_user in chatdata.participants:
+        for msg in inp.chatdata.messages:
+            if msg.from_user in inp.chatdata.participants:
                 users_with_msg.add(msg.from_user)
-        for uid, user in chatdata.participants.items():
+        for uid in users_with_msg:
+            user_display_name = (inp.chatdata.participants[uid]).displayName
```
vingkan, line 27: FYI: No need for parentheses here.

vingkan (approved):
Excellent! Great work, Chris. Very thorough PR.

PR #328, opened 2022-07-25
Create story cases

gracesopha:
-create two user case
-create three user case
-set up for showing blocked user


vingkan (approved):
Great job, Grace! Thanks for adding these test cases.

sprihajha (approved):
Awesome work! 🎉 
Looks good to me 🚀 @gracesopha

PR #329, opened 2022-07-25
Added renderer parent class

chunter3:


pipeline/notifications/render_check_message.py
```diff
@@ -0,0 +1,28 @@
...
+
+
+class NewMessageRender(RenderNotification):
+    def __init__(self):
+        pass
+
+    def render(self, inp: NotificationInfo) -> str:
+
+        # if the notification type doesn't match the type of render, return immediately saying as much
+        if inp.notification_type.value != "check_message":
```
vingkan, line 12: You can compare the enum here. Also, raise an error instead of returning a string, otherwise the caller might consider the string to be a valid message!

```python
# You can move this exception to a file like `pipeline/notification/exceptions.py`
# or put it in `pipeline/notifications/renderer.py` since it will be used by other renderers
class InvalidNotificationTypeException(Exception):
  pass

if inp.notification_type != NotificationType.CHECK_MESSAGE:
  raise InvalidNotificationTypeException("This renderer only supports CHECK_MESSAGE notifications.")
```

pipeline/notifications/render_check_message.py
```diff
@@ -0,0 +1,28 @@
...
+        elif len(inp.messagers) > 2:
+            x = 0
+            for msg in inp.messagers:
+                if x == len(inp.messagers) - 1:
+                    str_msg_lst += f"and {msg}."
+                    break
+                str_msg_lst += f"{msg}, "
+                x += 1
+        else:
+            str_msg_lst = "".join(inp.messagers) + "."
```
vingkan, line 26: You can move this into a helper function so that (a) it could be reused and (b) the renderer code is more readable. Then it would look like this:

```python
messagers_list = get_messagers_names(inp.messager)
text = f"..."
return text
```

You can put the helper function in this file for now since it is currently only relevant to this use case. If others have a need for it in the future, they can refactor it by moving it to common code.

pipeline/notifications/render_check_message.py
```diff
@@ -0,0 +1,28 @@
+from pipeline.notifications.renderer import RenderNotification
+from pipeline.types import NotificationInfo
+
+
+class NewMessageRender(RenderNotification):
+    def __init__(self):
+        pass
+
+    def render(self, inp: NotificationInfo) -> str:
```
vingkan, line 9: I would call this argument `info` or `data` instead of `inp`, to avoid confusion with `NotificationInfo`.

pipeline/notifications/render_check_message_test.py
```diff
@@ -0,0 +1,51 @@
...
+
+    test_notif_info = NotificationInfo(
+        recipient=User(uid="A", displayName="Ayman", email="ayman@iit.edu"),
+        notification_type=NotificationType.RATE_MATCH,
+        messagers=["Chris"],
+    )
+    test_render = NewMessageRender()
+    actual = test_render.render(inp=test_notif_info)
+    expected = "Renderer does not support this type of notification."
+    assert actual == expected
```
vingkan, line 51: Good job adding a test for this! After you change the render to raise an exception if the wrong notification type is passed in, you can use `pytest.raises()` to test that the exception of the correct type was raised.

This StackOverflow post gives a good example: https://stackoverflow.com/a/29855337

pipeline/notifications/renderer.py
```diff
@@ -0,0 +1,23 @@
...
+      the associated child classes, which are the implementation of the
+      other renders for the notification types
+
+    * A render, similiar to a notifier, only works on one notification at a time
+    """
+
+    def __init__(self, name: str):
+        self.name = name
+
+    def render(self, inp: NotificationInfo) -> str:
```
vingkan, line 17: Same comment here about renaming `inp` to `info` or `data`.

pipeline/notifications/__init__.py
```diff
@@ -1,2 +1,4 @@
 from pipeline.notifications.check_message import *
 from pipeline.notifications.notifier import *
+from pipeline.notifications.render_check_message import *
```
vingkan, line 3: There's a slight discrepancy between your class names and your file names. Maybe rename the renderer and notifier files to match the class name?

vingkan (changes requested):
Great start, Chris! Overall this looks great, just left a few comments to revise and then will approve.

pipeline/notifications/__init__.py
chunter3: Yeah I see what you mean. In fact, I'll go ahead and rename all of the notifications files so that they're more consistent in their naming.

pipeline/notifications/new_message_render.py
```diff
@@ -0,0 +1,41 @@
...
+    RenderNotification,
+)
+from pipeline.types import NotificationInfo, NotificationType
+
+
+class NewMessageRender(RenderNotification):
+    def __init__(self):
+        pass
+
+    def format_messager_names(self, inp: Set[str]) -> str:
```
vingkan, line 14: Since this method does not use any state in the class (it never accesses `self`) and it is not part of the external functions of this class, it does not have to be an instance method.

So, you could move it to be its own function in this file file and remove `self`.

There is also a decorator called `@classmethod` that marks class methods that don't rely on any state. But that would only make sense if this were an external function that we expect users of this class to have access to. In this case, it is a function used only internally, so we would take it out of the class entirely.

pipeline/notifications/new_message_render_test.py
```diff
@@ -0,0 +1,59 @@
...
+    )
+    test_renderer = NewMessageRender()
+    with pytest.raises(
+        InvalidNotificationTypeException, match="only supports CHECK_MESSAGE"
+    ) as exc_info:
+        actual = test_renderer.render(info=test_notif_info)
+    assert (
+        exc_info.value.args[0]
+        == "This renderer only supports CHECK_MESSAGE notifications."
+    )
```
vingkan, line 59: Looks good! In the future, you don't even have to check the exception text, usually matching on the type of exception is sufficient. That's what we create new custom exception types whose name describes the problem.

vingkan (approved):
Excellent work, Chris! Looks good to go.

PR #330, opened 2022-07-25
Fixes #90 Block User Functionality

### LINKED ISSUE #90: Chat App: Block Users
## Goal

Implement a feature in the chat menu that allows a user to block the other user(s) in the chat from sending them messages, in case they are being harassed.

## Definition of Done

- Users can access the feature from the chat menu
- When a user blocks another user, they will not see any messages from that user in any of their chats.
- Users can un-block or re-block another user at anytime.
- All participants in the chat should be able to see that a user has blocked another user.

## Starter Code

```
-- Write your styles here
frontend/src/app/chatmenu/BlockUsers.css
-- Implement your component here
frontend/src/app/chatmenu/BlockUsers.tsx
-- Write your test stories here
frontend/src/app/chatmenu/BlockUsers.stories.tsx
```

sprihajha:
- Added implementation to Block/Unblock user.
- Added support to fetch blocked users and style User Tile accordingly.


frontend/src/app/chatmenu/BlockUser.tsx
```diff
@@ -1,66 +1,79 @@
...
-                        onClick={doUnblock}
-                        icon={faUnlock}
-                        className="IconBeforeUnblock"
-                    />
-                    Unblock
-                </button>
-            </div>
+        <div className={isActive ? 'BlockedUser' : ''}>
+            <UserTile user={user} />
+            <Button
```
vingkan, line 49: Nice reuse of built-in components.

frontend/src/app/chatmenu/BlockUser.tsx
```diff
@@ -1,66 +1,79 @@
...
+    return blockedUsers.find((user: { uid: string }) => user?.uid === blockUid)?.blocked || false
 }
 
 // create user tile to display participants
-const UserTile = ({ user }) => {
-    const { uid, displayName, photoURL } = user
-    const alt = `Profile image for ${displayName}.`
+const CreateUserTile = ({ user, myUid }) => {
+    const blockUid = user?.uid
+    const [isActive, setActive] = useState(FetchBlockedUsers(myUid, blockUid))
```
vingkan, line 30: I think we should move this state out of the `CreateUserTile` functional component, otherwise we will send a request each time we render a user. Instead, we should fetch the information at the top-level component, then pass in whether or not the user is blocked into the user element.

frontend/src/app/chatmenu/BlockUser.tsx
```diff
@@ -1,66 +1,79 @@
...
+
+function FetchBlockedUsers(myUid: string, blockUid: string) {
+    const [res] = useBackendFetchJson({
+        route: `/chat/block/user/${myUid}`,
+        deps: [myUid],
+        isValid: myUid,
+    })
+    const blockedUsers = res?.blocks || []
+
+    return blockedUsers.find((user: { uid: string }) => user?.uid === blockUid)?.blocked || false
```
vingkan, line 24: Instead of using `find()` here, we could reduce the list into a dictionary of `uid` to `blocked`, then when we loop over the participants, we can pull the blocked status from the dictionary in constant time, with `false` as the fallback value if the `uid` is not found.

There are only a maximum of three users in participants, but there could be more users in the blocked list. Even though the total number is not likely to be high enough to cause a performance issue, we can still improve the runtime and demonstrate good frontend performance for an efficient user interface.

frontend/src/app/chatmenu/BlockUser.tsx
```diff
@@ -1,66 +1,79 @@
...
+        })
     }
 
-    const doUnblock = () => {
-        // TODO: Implement unblocking the user method
+    const doUnblock = async () => {
+        await fetchFromBackend({
+            route: `/chat/block/user/${myUid}/${blockUid}/false`,
+            options: { method: 'POST' },
+        })
```
vingkan, line 43: These functions can also be partially lifted out of the `CreateUserTile` functional component. Then, the user tile doesn't need to have the extra prop of `myUid`.

Instead, the user tile can take a function prop:
```ts
doBlock(blockUid: string, shouldBlock: bool) => void
```

And then the component can use that function as the onclick callback. The parent component will take the input values and use that to send the actual request.

This also separates the component from the request concern, leaving only the event handling.

vingkan (changes requested):
Excellent start! This is a good example of a frontend feature with a tricky relationship between the outer and inner components.

I left feedback on the separation of concerns between parent and child component, lifting state up, passing in a partial function instead of a dependent prop, and an optional performance improvement.

frontend/src/app/chatmenu/BlockUser.tsx
```diff
@@ -1,66 +1,81 @@
...
+        {}
+    )
+}
 
-    const doBlock = () => {
-        // TODO: Implement blocking the user method
-    }
+async function updateBlock(blockUid: string, shouldBlock: boolean) {
+    await fetchFromBackend({
+        route: `/chat/block/user/${this?.BlockUserProps?.myUid}/${blockUid}/${shouldBlock}`,
```
vingkan, line 35: Ah, this is why the block results don't get saved. I don't think there is a problem with the `useEffect()`, instead the problem is that we send an `undefined` value for the `from_uid` in the endpoint, so there is no way to know who is blocking this user.

The reason is that `this` does not have access to the properties we are using here.

There are a few different wants to handle this:

Option A: Revise the `updateBlock()` function to take `myUid` as an input, which means that we then have to pass `myUid` into the `CreateUserTile` component. The downside of this is that now we are drilling a prop we shouldn't really need.

Option B: Pass an `updateBlock()` function to the `CreateUserTile` component that takes only `blockUid` and `shouldBlock` as arguments, then use partial application to create a function that already has `myUid` set, then pass that partially applied function to the component. This might add more indirection, but it limits what the inner component needs to provide.

In the outer component, it would look something like this:

```jsx
async function blockUser(myUid, blockUid, updateBlock ) { ... }

// Renamed from CreateUserTile
function BlockUserTile({ user, value, submitBlock }) {
  const blockUid = user?.uid
  const [blocked, setIsBlocked] = useState(value)

  const blockHandler = () => {
    const newValue = !value
    submitBlock(blockUid, newValue)
    setIsBlocked(newValue)
  }

  return (...)
}

function OuterComponent({ myUid }) {
  // Partial application turns a function that needs three inputs into one that needs only two
  const updateBlock = (blockUid, shouldBlock) => blockUser(myUid, blockUid, shouldBlock)
  return (
    ...
    <BlockUserTile user={user} value={value} updateBlock={updateBlock} />
  )
}
```

vingkan (changes requested):

frontend/src/app/chatmenu/BlockUser.tsx
```diff
@@ -1,66 +1,83 @@
...
-    }
+async function updateBlock(myUid: string, blockUid: string, shouldBlock: boolean) {
+    await fetchFromBackend({
+        route: `/chat/block/user/${myUid}/${blockUid}/${shouldBlock}`,
+        options: { method: 'POST' },
+    })
+}
+
+// create user tile to display participants
+const CreateUserTile = ({ user, value, submitBlock }) => {
```
vingkan, line 41: Optional: Maybe we can rename this component to `BlockUserTile` so that it is a noun instead of a verb?

frontend/src/app/chatmenu/BlockUser.tsx
```diff
@@ -1,66 +1,83 @@
...
         </div>
     )
 }
 
 // show the users within the chat
-const BlockUserInner = ({ participants }: BlockUserProps) => {
+const BlockUserInner = ({ participants, myUid }: BlockUserProps) => {
+    const blockedUsers = useFetchBlockedUsers(myUid)
+    const submitBlock = (blockUid: string, shouldBlock: boolean) =>
+        updateBlock(myUid, blockUid, shouldBlock)
```
vingkan, line 70: Nice! The partial application here looks clean.

vingkan (approved):
Great work, Spriha! Looks good to merge.

Issue #259, opened 2022-07-12
Remove references to generator-specific intent and interest fields


vingkan:
## Goal

A final clean up task for #258. After the intent and interest generators are implemented, refactor them so that they only use the shared match metadata fields for intents and interests. Then delete the old metadata fields.

## Definition of Done

- [ ] No code uses the old, generator-specific intent and interest fields
- [ ] TypeScript definitions in the app are in sync with the pipeline dataclasses

PR #335, opened 2022-07-26
Minor Changes to First Similar Intent Solution

lwilliams717:
Changed a few things about the first solution based on mentor comments:

1. Remove score from the match metadata 
2. Changing the user variable names in for loops
3. Added a new test case for the min common intents to be implemented later. For now the test case is commented out.


pipeline/matching/generators/similar_intents.py
```diff
@@ -1,18 +1,91 @@
...
+                        # checks if the users have been matched, prevents duplicates
+                        if key in all_matches:
+                            # additional check if the users have NOT matched on this intent and appends it to the matchingIntents
+                            if (
+                                intent_match
+                                not in all_matches[key].metadata.matchingIntents
+                            ):
+                                all_matches[
+                                    key
+                                ].metadata.matchingIntents.append(intent_match)
```
vingkan, line 79: Please use the new `match.metadata.intents` field instead of `matchingIntents`. Refer to #259 and #327 for more information on this change.

lwilliams717: Okay, I will update this now!

vingkan:
Closing due to conflicts.

PR #336, opened 2022-07-27
Notifications per Users

azuhdz7:


pipeline/utils/notifications_test.py
```diff
@@ -0,0 +1,43 @@
+import pytest
+
+from pipeline.types import NotificationInfo, NotificationType
+from pipeline.utils.notifications import notifications_per_user
+
+
+def test_return_lst_1_notification():
+    notifications = [
+        NotificationInfo(
+            recipient="A",
```
vingkan, line 10: The reason you got stuck on the recipient issue is because the type of this property is `User`, not a string or user ID like you have used before in other classes.

Check out the type definition:

https://github.com/scarletstudio/butterfly/blob/bf74db6426ead1e0fbd13d7254c388d56c091d84/pipeline/types/notification.py#L27-L34

Instead, you can create a recipient like this (make share to import the `User` type):

```python
notifications = [
  NotificationInfo(
    recipient=User(uid="A"),
    notification_type=NotificationType.CHECK_MATCH
  )
]
```

You can also leave out `messagers` since it is not affected by the function you are writing, but if you were setting that property, it would have to be a list of strings, per the type definition.

pipeline/utils/notifications_test.py
```diff
@@ -0,0 +1,43 @@
...
+            notification_type=NotificationType.CHECK_MATCH,
+            messagers="B",
+        )
+        # NotificationInfo(recipient="B", notification_type=NotificationType.CHECK_MATCH, messagers="A")
+    ]
+    actual = notifications_per_user(notifications)
+    expected = [
+        NotificationInfo(
+            recipient="A",
+            notification_type=NotificationType.CHECK_MATCH,
```
vingkan, line 20: Good job using the enums!

pipeline/utils/notifications_test.py
```diff
@@ -0,0 +1,43 @@
...
+            messagers="B",
+        )
+        # NotificationInfo(recipient="B", notification_type=NotificationType.CHECK_MATCH, messagers="A")
+    ]
+
+    assert actual == expected
+
+
+# def test_return_lst_with_double_notifications():
```
vingkan, line 29: Excellent, this is a good test case that tests for excluding repeat notifications!

vingkan (changes requested):
Good start! I commented on the `recipient` issue, but this is on the right track.

vingkan (approved):
Well done, Azu!

PR #341, opened 2022-07-28
Improved rare intents match generation

nelmajdoub:


pipeline/matching/generators/rare_intents.py
```diff
@@ -1,18 +1,138 @@
...
+                                )
+                            ],
+                        ),
+                    )
+                    user_one_in_match = seeker_user.uid
+                    user_two_in_match = giver_user.uid
+
+                    checking_same_user = False
+
+                    if int(user_one_in_match) > int(user_two_in_match):
```
arushirai1, line 116: helpful to push this into a separate function and also leave a comment about what this if-else snippet is trying to achieve. Could also do this in a single line using the sort function - `''.join([user_one_in_match, user_two_in_match].sort())`

arushirai1 (approved):

PR #349, opened 2022-07-28
Creating notifications file in pipeline/flows

chunter3:


pipeline/flows/notifications.py
```diff
@@ -0,0 +1,93 @@
...
+    logger.info("Generating notifications")
+    # refactor to account for additional notification types once they're implemented
+    notif_info_lst = notifier.get_notifications(inp=inp)
+    logger.info(
+        f"Done. Number of notifications generated: {len(notif_info_lst)}"
+    )
+    return notif_info_lst
+
+
+def notif_hierarchy(notifs: List[NotificationInfo]) -> List[NotificationInfo]:
```
chunter3, line 29: What do you think about how I went about implementing the hierarchy? Do you agree with what I've decided the hierarchy should be? Do you believe I should do something different? I'm open to any suggestions to improve the hierarchy function in any way.

pipeline/flows/notifications.py
```diff
@@ -0,0 +1,93 @@
...
+    Extract chat data for the community in current release
+
+     * Will contain messages and information about the match
+
+     * David is working on implementing this; add it to flow once he's finished
+    """
+    # TODO: add chat data extraction
+
+    # Extract users in the community
+    df_users = extract_users(db, param_community)
```
chunter3, line 82: Is there a reason we need to get all the users in the community? Because I really can't think of one for this particular flow. The only users that matter are the ones we need to notify, and we have those users in a list of NotificationInfo objects (i.e. List[NotificationInfo])

pipeline/flows/notifications.py
```diff
@@ -0,0 +1,93 @@
...
+    logger = prefect.context.get("logger")
+    # add other types of notifiers once they are implemented
+    notifier = NewMessageNotifier()
+    logger.info("Generating notifications")
+    # refactor to account for additional notification types once they're implemented
+    notif_info_lst = notifier.get_notifications(inp=inp)
+    logger.info(
+        f"Done. Number of notifications generated: {len(notif_info_lst)}"
+    )
+    return notif_info_lst
```
vingkan, line 26: Looks good! You could start by creating a list of notifiers, even if right now, it will only have one notifier in it. Then you can adapt your logic to loop through the notifiers and generate all the notifications.

pipeline/flows/notifications.py
vingkan, line 82: Everything that is part of the flow should be indented within the `with Flow(...) as flow:` block, not just the parameters.

vingkan: The chat data does not contain the actual user objects, it only contains their user IDs. So when @dfg-13 and @notedwin extract the chat data, they will either need to extract the users or take in a list of users that have already been extracted, so that they can add them to the chat data output.

Either way works, let's see what they decide to do! For now, you can leave out the extract users and add it back if they need it.

pipeline/flows/notifications.py
vingkan: This looks fine! Let's leave it how it is because I don't want you to focus on this one too much.

@azuhdz7 is working on their own implementation of this, what you have works for now, but they have another approach and tests that we can have you review.

pipeline/flows/notifications.py
```diff
@@ -0,0 +1,93 @@
...
+    # Transform users from dataframe to user objects
+    users_w_profile = convert_users_from_df(df_users)
+
+    # check hierarchy of notifications
+    chatdata = ChatData(release_timestamp=datetime.datetime.now())
+    inp = NotificationInput(
+        chatdata=chatdata, run_datetime=datetime.datetime.now()
+    )
+    notifs = get_users_to_notify(inp=inp)
+    final_notifications = notif_hierarchy(notifs=notifs)
```
vingkan, line 93: A single notifier only operates on one chat, but the notifications pipeline should operate on many chats. I would recommend adding a task that takes in a list of ChatData and generates notifications for each one. The reason for a new task is that we prefer not to have loops in the flow code itself, so that Prefect can figure out the task ordering.

vingkan, line 93: Where will you add the notification renderers? And how will we know which renderer to use for each notification type?

pipeline/flows/notifications.py
```diff
@@ -0,0 +1,93 @@
...
+        secret_admin_credentials = PrefectSecret("admin_credentials")
+        db = initialize_firebase_for_prefect(
+            secret_database_url, secret_admin_credentials
+        )
+    """
+    Extract chat data for the community in current release
+
+     * Will contain messages and information about the match
+
+     * David is working on implementing this; add it to flow once he's finished
```
vingkan, line 77: While you wait for David, you can also create some fake chat data for this pipeline. It's okay to just write it as a local variable like:

```python
fake_extracted_chats = [
  ChatData(...),
  ChatData(...),
  ChatData(...)
]
```

You might need to write a small Prefect task that returns the fake data, so that you can use that task as a placeholder for the extract task. In Prefect, a task does not return a value, it returns an output, that is then turned into a value by Prefect if that output is used, so that's why you probably won't be able to just make a variable and pass it into your other tasks.

vingkan (changes requested):
Looking good, Chris! I left comments and responded to your questions.

Also suggested a way for you to not be blocked by David's extract work and to leave your hierarchy code as-is, since Azu is making good progress on that and I want you to focus on the other aspects of the pipeline.

You should be able to test running the pipeline with `run flow notifications`.

pipeline/flows/notifications.py
chunter3: Understood.

pipeline/flows/notifications.py
chunter3: I see. Excellent. I'm interested to see @azuhdz7 implementation. I'm sure it'll be more effective than I have at the moment.

pipeline/flows/notifications.py
chunter3: Isn't this more or less accomplished with the get_users_to_notify() prefect task? Should I refactor it to have the input be a list of ChatData and have it be a generator instead (i.e yield a List[NotificationInfo] for each ChatData in the list)?

chunter3: I've decided to refactor the existing prefect task (I also renamed it to get_notifications())

pipeline/flows/notifications.py
vingkan: Awesome, we will do that in a follow-up PR after #366 merges!

pipeline/flows/notifications.py
```diff
@@ -0,0 +1,144 @@
...
+def render_notifications(
+    notifs: List[NotificationInfo],
+) -> List[str]:
+    logger = prefect.context.get("logger")
+    output_lst = []
+    for notif in notifs:
+        renderer = RENDERERS.get(notif.notification_type)
+        if renderer:
+            output_str = renderer.render(notif)
+            logger.info(output_str)
```
vingkan, line 110: In the future, we need to add a `NotificationOutput` dataclass that includes the recipient email, subject, and body.

pipeline/flows/notifications.py
```diff
@@ -0,0 +1,144 @@
...
+        )
+        """
+        Extract chat data for the community in current release
+
+        * Will contain messages and information about the match
+
+        * David is working on implementing this; add it to flow once he's finished
+        """
+        # TODO: add chat data extraction
+        pseudo_chatadata = return_pseudo_chatdata(param_community)
```
vingkan, line 137: Either David or Chris will hook this up when #365 merges.

vingkan (approved):
Great work, Chris!

PR #354, opened 2022-07-29
Kept the messages in the search bar

NidaAkkiswala:
Added the following enhancements to the Search Messages feature:

1. **Retain search term:** To avoid losing context about the search, retained the search term within the search `Textarea` element with a cross, an icon to clear out the previous search easily. 

2. **`cleanText` Helper function:** Added `cleanText(value)` in `frontend/src/app/chatmenu/SearchMessages.stories.tsx` file. This helper function addresses the following issues with the search term and messages being filtered:
- Returning `''` if there is no given value
- Converting search terms and messages to lower case
- Trimming whitespace

### DIFF
--- a/frontend/src/app/chatmenu/SearchMessages.css
+++ b/frontend/src/app/chatmenu/SearchMessages.css
@@ -12,7 +12,6 @@
 }
 
 .Search .Input {
-    position: relative;
     --input-width: calc(80% - var(--send-width));
     min-width: var(--input-width);
     border-top-left-radius: var(--chat-composer-radius);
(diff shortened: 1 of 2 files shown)


vingkan (commented):
This looks good to me! I will leave it to @sprihajha to approve in case there are other edits she recommends. Nice job, @NidaAkkiswala!

sprihajha:
@vingkan Just waiting on the clear search text icon to be added. Otherwise, everything looks good!

sprihajha (approved):
All the requested changes are in! Looks good 🚀

PR #355, opened 2022-07-31
Vocable word guessing game

BoominHutJr:
Outlined functions, input, and output for word guessing game


backend/games/wordGuesser/ClassicWordle.py
```diff
@@ -0,0 +1,33 @@
+from typing import List
```
vingkan, line 1: Can you make the folder name `word_guesser` instead of `wordGuesser`? Our Python codebase uses snake_case for module/function names/variables and camelCase for class names. `ClassicWordle` is fine for the class name!

It would also be more consistent with the codebase as a whole if your functions and variables were also snake_case, but it's not the most important thing.

backend/games/wordGuesser/ClassicWordle.py
```diff
@@ -0,0 +1,33 @@
...
+
+
+class ClassicWordle:
+    # maximum number of guesses
+    MAX_GUESSES = 6
+    # represents the state of the game, index 0 will hold the "goal word"
+    gameBoard = List[str]
```
vingkan, line 8: FYI `gameBoard` is a class-variable, not and instance-variable, which means all objects of this class will mutate the same variable. We want to be able to host multiple games at the same time, so this should be changed to an instance variable.

More details here: https://www.tutorialsteacher.com/articles/class-attributes-vs-instance-attributes-in-python

`MAX_GUESSES` is fine as a class variable.

backend/games/wordGuesser/ClassicWordle.py
```diff
@@ -0,0 +1,33 @@
...
+
+    # clears any previous activity and randomizes a word to initialize the list
+    def startGame(self) -> List[str]:
+        pass
+
+    # clears all previous progress and calls checkWin() to determine final results
+    # will return a victory or failure string
+    def endGame(self) -> str:
+        pass
```
vingkan, line 17: If `startGame()` already clears the game state, why does `endGame()` need to?

And why have an `endGame()` at all? Could the win/loss state be returned with `trackProgress()`?

backend/games/wordGuesser/ClassicWordle.py
```diff
@@ -0,0 +1,33 @@
...
+    # returns list of letters correctly guessed
+    # capital= right letter, right space
+    # lower = right letter wrong space
+    def trackProgress(self) -> List[str]:
+        pass
+
+    # checks to see if win/lose conditions have been satisfied
+    # meant to run after each guess, and after MAX_GUESS has been reached
+    def checkWin(self) -> bool:
+        pass
```
vingkan, line 33: Thanks for writing this!

Base on this, this is how I am planning to write the UI (frontend) and API (backend) flow that calls your game class:

1. User starts a game
  - UI starts listening to the database for their game data, although at first there is nothing
  -  UI sends a request to start a game
  - API calls your class to start a game
  - API assigns the game class instance to an ID for the chat
  - API saves an initial game state to the database
  - UI listens to this database update and displays the game board
2. User submits a guess
  - UI sends a request with the guess
  - API calls your class to submit the guess
  - API calls your class to check the win condition
  - API calls your class to track the progress
  - API saves the game state data to the database
  - UI listens to this database update and displays the game state
 3. The above continues until a game end condition is reached

Your code will fit into this flow. But one concern is that if the API needs to be restarted, we will lose all game class objects and the user will not be able to continue playing their game. In the future, you could consider designing a game class that is "stateless" or functional, in that it takes in any state data given to it, and can resume the game from there. But for this last week of the internship, your current approach should work.

vingkan (changes requested):
Requested a few small changes, then will approve. Good start @BoominHutJr!

backend/games/wordGuesser/ClassicWordle.py
BoominHutJr: My original idea was for start_game() to clear previous game data as a precaution, such that starting a new game will not be affected by any previous game data. Ideally, end_game() would be called within start_game().

having this as a separate function also allows the users to terminate the game early if they decide to quit and find out what the word was, all without starting a new game

vingkan: That makes sense!

vingkan (changes requested):
Left another comment on tracking the progress!

backend/games/wordGuesser/ClassicWordle.py
```diff
@@ -0,0 +1,33 @@
...
+    # maximum number of guesses
+    MAX_GUESSES = 6
+    # represents the state of the game, index 0 will hold the "goal word"
+    gameBoard = List[str]
+
+    # clears any previous activity and randomizes a word to initialize the list
+    def startGame(self) -> List[str]:
+        pass
+
+    # clears all previous progress and calls checkWin() to determine final results
```
erikqu, line 14: Please change these comments to block comments

backend/games/wordGuesser/ClassicWordle.py
```diff
@@ -0,0 +1,33 @@
...
+    # performs all logic checks on a word to determine if it is valid
+    # takes a word and updates the board based on
+    def guessWordle(self, guess: str) -> List[str]:
+        pass
+
+    # returns list of letters correctly guessed
+    # capital= right letter, right space
+    # lower = right letter wrong space
+    def trackProgress(self) -> List[str]:
+        pass
```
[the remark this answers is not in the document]
BoominHutJr: I think the first guess dict would be easier to work with so id prefer the list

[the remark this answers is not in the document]
vingkan: Sounds good, let's do that!

backend/games/word_guesser/ClassicWordle.py
```diff
@@ -0,0 +1,137 @@
...
+    def start_game(self) -> List[str]:
+        # clear past game data
+        guesses = {"guess1": "", "results1": []}
+        game_board = [""]
+        goal_word = str
+        guess_count = 0
+
+        # randomize next word
+        new_word_index = randint(0, 19)  # nosec
+        goal_word = WORD_LIST[new_word_index]
```
vingkan, line 48: This will cause a `mypy` error because it does not find `WORD_LIST` in scope. Since you have `WORD_LIST` as a class variable, try to access it by `ClassicWordle.WORD_LIST`.

Or you can make it a global variable in this file, not just the class.

backend/games/word_guesser/ClassicWordle.py
```diff
@@ -0,0 +1,137 @@
...
+
+    # clears all previous progress and calls checkWin() to determine final results
+    # will return a victory or failure string
+    def end_game(self) -> str:
+        guesses = {"guess1": "", "results1": []}
+        game_board = List[str]
+        goal_word = ""
+        guess_count = 0
+
+        if check_win():
```
vingkan, line 59: Here you have to call `self.check_win()` to access the instance method.

backend/games/word_guesser/ClassicWordle.py
```diff
@@ -0,0 +1,137 @@
...
+        -list will determine correctness of each letter in guess in the form of :
+            "correct", "in_word", "not_in_word"
+        -updates guesses dict
+    """
+
+    def track_progress(self) -> Dict:
+        progress = {}
+        tmp_keys = ""
+        tmp_results = []
+        for k, v in guesses.items():
```
vingkan, line 75: The variable `guesses` is not in scope. It looks like you want the instance variable, so you should use `self.guesses`. That means you also need to update your constructor to set the instance variables:

```python
    def __init__(self):
        ...
        self.guesses = {"guess1": "", "results": []}
        ...
```

backend/games/word_guesser/ClassicWordle.py
```diff
@@ -0,0 +1,137 @@
...
+        "rouse",
+        "fraud",
+        "atone",
+        "raise",
+        "minor",
+    ]
+
+    def __init__(self):
+        # represents the state of the game, index 0 will hold the "goal word"
+        game_board = List[str]
```
vingkan, line 33: This is not a valid initialization, the type hint and initial value need to be defined separately, because the type is just a hint, not a value:

```python
game_board: List[str] = []
```

This shows how we add a colon after the variable name to add the type hint, then put the value after equals sign.

vingkan (changes requested):

backend/games/word_guesser/WordGuesser.py
```diff
@@ -0,0 +1,137 @@
+from random import randint
+from typing import Dict, List
+
+from String import isalpha
+
+MAX_GUESSES = 6
+WORD_LIST = [
```
erikqu, line 7: We should put this word list in a separate file

backend/games/word_guesser/WordGuesser.py
```diff
@@ -0,0 +1,137 @@
...
+    "atone",
+    "raise",
+    "minor",
+]
+
+
+class WordGuesser:
+    def __init__(self):
+        # represents the state of the game, index 0 will hold the "goal word"
+        self.game_board: List[str] = [""]
```
erikqu, line 34: I like the implied types here!

backend/games/word_guesser/WordGuesser.py
```diff
@@ -0,0 +1,137 @@
...
+    # clears any previous activity and randomizes a word to initialize the list
+    def start_game(self):
+        # clear past game data
+        self.guesses = {"guess1": "", "results1": []}
+        self.game_board: List[str] = [""]
+        self.goal_word = ""
+        self.guess_count = 0
+
+        # randomize next word
+        new_word_index = randint(0, 19)  # nosec
```
erikqu, line 48: Lets make this dynamic to the size of the list above

backend/games/word_guesser/WordGuesser.py
```diff
@@ -0,0 +1,137 @@
...
+        goal_word = WORD_LIST[new_word_index]
+
+    # clears all previous progress and calls checkWin() to determine final results
+    # will return a victory or failure string
+    def end_game(self) -> str:
+        self.guesses = {"guess1": "", "results1": []}
```
erikqu, line 54: These lines seem common, we should put them in a method

backend/games/word_guesser/WordGuesser.py
```diff
@@ -0,0 +1,137 @@
...
+        elif self.guess_count >= MAX_GUESSES:
+            return False
+        return False
+
+    # performs all logic checks on a word to determine if it is valid
+    def validate_guess(self, guess) -> bool:
+        if len(guess) != 5:
+            print("Word length must be 5.")
+            return False
+        if not isalpha():
```
erikqu, line 98: does isalpha need an argument here?

backend/games/word_guesser/WordGuesser.py
```diff
@@ -0,0 +1,137 @@
...
+        self.game_board: List[str] = [""]
+        self.guesses = {"guess1": "", "results": []}
+        self.goal_word = ""
+        self.guess_count = 0
+
+    # clears any previous activity and randomizes a word to initialize the list
+    def start_game(self):
+        # clear past game data
+        self.guesses = {"guess1": "", "results1": []}
+        self.game_board: List[str] = [""]
```
erikqu, line 43: This isn't really a game board, is it?

backend/games/word_guesser/WordGuesser.py
```diff
@@ -0,0 +1,137 @@
...
+                    char.lower() is not self.goal_word.split()[i]
+                ):
+                    curr_results[i] = "in_word"
+                else:
+                    curr_results[i] = "not_in_word"
+
+            # track all guesses by labling them guess1, guess2, etc and inputing them as new dict entries
+            key = "guess" + str(self.guess_count)
+            value = "results" + str(self.guess_count)
+            self.guesses[key] = guess
```
erikqu, line 126: I think this is a little clever. Would it make more sense to have a list of hash maps here instead, so we can get rid of the numbers in the keys here for consistency?

erikqu (changes requested):

backend/games/word_guesser/WordGuesser.py
BoominHutJr: Will do, ill save this one for last

backend/games/word_guesser/WordGuesser_test.py
```diff
@@ -0,0 +1,73 @@
...
+        "guess": "minor",
+        "results": [
+            "correct",
+            "not_in_word",
+            "not_in_word",
+            "in_word",
+            "not_in_word",
+        ],
+    }
+    assert game.guesses["guess"] == tmp_guess["guess"]
```
vingkan, line 59: How does this test case pass? If the goal word is chosen randomly, how do you make sure the guess `"minor"` has the same results?

Ideally, you would have the process for choosing a word separate from the game logic, that way you do not have randomness in a component that you are testing for deterministic logic. And you can test the random word chooser separately, even though it is random, it will adhere to some testable properties, like how you asserted that the goal word is in the list of options.

The next best thing is to be able to override the goal word when instantiating the game class. If a goal word is specified in the constructor, then use that, otherwise pick one at random.

This will allow you to write deterministic test cases for your game class and test specific edge cases.

vingkan (changes requested):
Great progress, Ayman! I left some advice on making the tests deterministic.

backend/games/word_guesser/WordGuesser_test.py
BoominHutJr: - changed so that the default goal word is "basic" 
- start_game() is the function that randomizes the goal word, so it is not called in the tests
- start_game() is tested separately in test_start_game()

backend/games/word_guesser/WordGuesser.py
BoominHutJr: It represents only the previous guesses and is meant to be used more internally than on the frontend

backend/games/word_guesser/WordGuesser.py
BoominHutJr: it did, thank you

backend/games/wordGuesser/ClassicWordle.py
BoominHutJr: All comments have been filtered and cleaned

vingkan (approved):
Great work, Ayman!

@erikqu and I have feedback on the implementation and testing, but you can make those changes in a future PR. Let's merge this for now and then you can get started on revisions and we can find time to pair program on integrating this into the app.

backend/games/word_guesser/WordGuesser.py
```diff
@@ -0,0 +1,112 @@
+from random import randint
+from typing import Dict, List
+
+MAX_GUESSES = 6
+
+# initialize valid guesses and possible solutions
+SOLUTIONS: str
+VALID_WORDS: str
+
+solutions_file = open("backend/games/word_guesser/WordGuesser_Solutions.txt")
```
erikqu, line 10: Should we have exception handling here if this fails? Then the game won't be able to start

backend/games/word_guesser/WordGuesser.py
```diff
@@ -0,0 +1,112 @@
...
+    def track_progress(self) -> Dict:
+        tmp_keys = ""
+        tmp_results = []
+        for k, v in self.guesses.items():
+            if "guess" in k:
+                tmp_keys = v
+            if "results" in k:
+                tmp_results = v
+            self.progress[tmp_keys] = tmp_results
+        print(self.progress)
```
erikqu, line 57: We shouldn't have print statements in production code unless we're logging something important

backend/games/word_guesser/WordGuesser.py
```diff
@@ -0,0 +1,112 @@
+from random import randint
+from typing import Dict, List
+
+MAX_GUESSES = 6
+
+# initialize valid guesses and possible solutions
+SOLUTIONS: str
+VALID_WORDS: str
```
erikqu, line 8: Thanks for using uppercase and underscores here.

backend/games/word_guesser/WordGuesser.py
```diff
@@ -0,0 +1,112 @@
...
+        return True
+
+    def guess_word(self, guess: str) -> Dict:
+        if self.validate_guess(guess):
+            curr_results: List[str] = []
+            goal = self.goal_word
+            i = 0
+            for char in guess:
+                if char is goal[i]:
+                    curr_results.append("correct")
```
erikqu, line 85: maybe the correct/not_correct/etc should be an enum rather than strings? 

That would make the code more understandable and the actual enum can be integers rather than strings without giving up readability

backend/games/word_guesser/WordGuesser.py
```diff
@@ -0,0 +1,112 @@
...
+                elif (char in goal) and (char is not goal[i]):
+                    curr_results.append("in_word")
+                else:
+                    curr_results.append("not_in_word")
+                i += 1
+
+            # note that the previous guess is overwritten
+            self.guesses["guess"] = guess
```
erikqu, line 93: we should put these commonly used string names above since they are commonly used

erikqu: (applies everywhere)

backend/games/word_guesser/WordGuesser.py
```diff
@@ -0,0 +1,112 @@
...
+            self.game_board.append(self.guesses["guess"])
+            self.guess_count += 1
+
+            self.track_progress()
+
+            if self.guess_count == MAX_GUESSES:
+                self.end_game()
+        else:
+            self.guesses["guess"] = "INVALID"
+            self.guesses["results"] = [
```
erikqu, line 104: do we necessarily need to store all the guesses or can we have a variable that keeps track of the amount of guesses made and just store the last guess?

vingkan: I actually would prefer that we keep track of all of the guesses, right now this code is only keeping the last guess and the count of guesses.

We want to be able to prevent the user from guessing the same word multiple times.

However, you could also take in the previous guesses as an input, that way the database remains the source of truth for the game state and your class can receive the previous state to then update with the incoming guess. I can explain more on our call, since this is a more complicated design pattern than having a single stateful class.

backend/games/word_guesser/WordGuesser_test.py
```diff
@@ -0,0 +1,160 @@
+import pytest
+
+from backend.games.word_guesser.WordGuesser import WordGuesser
+
+SOLUTIONS: str
+
+solutions_file = open("backend/games/word_guesser/WordGuesser_Solutions.txt")
+guesses_file = open("backend/games/word_guesser/WordGuesser_ValidGuesses.txt")
```
vingkan, line 8: How did you get these lists? We should attribute the source of the list, at least in the source code.

backend/games/word_guesser/WordGuesser.py
```diff
@@ -0,0 +1,112 @@
...
+VALID_WORDS = guesses_file.read()
+VALID_WORDS_LIST = VALID_WORDS.splitlines()
+
+
+class WordGuesser:
+    def __init__(self):
+        self.guesses = {"guess": "", "results": []}
+        self.game_board: List[str] = []
+        self.progress: Dict = {}
+        self.goal_word = "basic"
```
vingkan, line 24: I was having trouble figuring out how your tests pass, then I realized that in your tests you are not calling `game.start_game()`, so then `"basic"` never gets replaced as the goal word.

That is clever, but it would be easier if we can specify any goal word so that we can have deterministic tests. You can add an optional argument to the constructor and use that word as the goal word if present. Then, when you call `game.start_game()`, if the word is not set, you can choose a random one, or if it is set, keep it.

backend/games/word_guesser/WordGuesser.py
```diff
@@ -0,0 +1,112 @@
...
+    def check_win(self) -> bool:
+        idx = len(self.game_board) - 1
+        if self.game_board[idx] == self.goal_word:
+            return True
+        elif self.guess_count > 6:
+            return False
+        return False
+
+    # performs all logic checks on a word to determine if it is valid
+    def validate_guess(self, guess) -> bool:
```
vingkan, line 69: I realized that your code only takes lowercase words, so I changed the UI to send lowercase words!

backend/games/word_guesser/WordGuesser.py
```diff
@@ -0,0 +1,112 @@
...
+SOLUTIONS: str
+VALID_WORDS: str
+
+solutions_file = open("backend/games/word_guesser/WordGuesser_Solutions.txt")
+guesses_file = open("backend/games/word_guesser/WordGuesser_ValidGuesses.txt")
+SOLUTIONS = solutions_file.read()
+SOLUTION_LIST = SOLUTIONS.splitlines()
+
+VALID_WORDS = guesses_file.read()
+VALID_WORDS_LIST = VALID_WORDS.splitlines()
```
vingkan, line 16: Might want to turn this into a set instead of a list for faster lookup. Same on the above line.

vingkan (commented):
Great work, Ayman! I left some feedback in addition to Erik's. You can also see #380 to see how we integrated your game into the app and API.

backend/games/word_guesser/WordGuesser.py
BoominHutJr: This makes more sense

backend/games/word_guesser/WordGuesser.py
BoominHutJr: Thanks, leftover print statement from debugging

backend/games/word_guesser/WordGuesser.py
BoominHutJr: Thanks for accomodating

backend/games/word_guesser/WordGuesser.py
BoominHutJr: self.progress stores the previous guesses and was intended to be used to give feedback to the user on what letters have have been used as well as how correct the letter was

backend/games/word_guesser/WordGuesser.py
BoominHutJr: enum implemented

backend/games/word_guesser/WordGuesser.py
```diff
@@ -0,0 +1,112 @@
...
+        if guess not in VALID_WORDS:
+            return False
+        return True
+
+    def guess_word(self, guess: str) -> Dict:
+        if self.validate_guess(guess):
+            curr_results: List[str] = []
+            goal = self.goal_word
+            i = 0
+            for char in guess:
```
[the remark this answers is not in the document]
BoominHutJr: enumerate implemented

backend/games/word_guesser/WordGuesser.py
BoominHutJr: id prefer to keep these as lists, as changing them to sets does not allow they to be easily indexed. Multiple "TypeError: 
'set' object is not subscriptable" errors arise

backend/games/word_guesser/WordGuesser_test.py
BoominHutJr: Pulled both lists from a public github repo

backend/games/word_guesser/WordGuesser.py
BoominHutJr: resolved through enum

backend/games/word_guesser/WordGuesser.py
erikqu: Ah ok

PR #356, opened 2022-08-01
Extract analytics data

arushirai1:
* Add capability to extract analytics from db (`extract_analytics from pipeline.extract.analytics`
* Add test for extract_analytics
* Add RawAnalytics type for output of `extract_analytics`


pipeline/extract/analytics_test.py
```diff
@@ -0,0 +1,28 @@
...
+    raw_analytics = {
+        "1": {
+            "data": {"user": "abc"},
+            "event_type": "view_chat_inbox",
+            "timestamp": 1659311549216,
+        }
+    }
+    mock_get_analytics = MagicMock(return_value=raw_analytics)
+    mock_db = MagicMock()
+    mock_db.reference("analytics/test").get = mock_get_analytics
```
vingkan, line 18: Great job setting up the unit test and mock data, very clean and easy to read!

vingkan, line 18: You can change this line in the test to:

```python
mock_db.reference("analytics").get = mock_get_analytics
```

Since there is just one analytics path in the database for the whole app.

I'm actually surprised that this passed, since the paths are different, but maybe pytest ignores the reference path, just returns the mock data for any call to `reference(...).get()`.

pipeline/extract/analytics.py
```diff
@@ -0,0 +1,10 @@
+from prefect import task
+
+from pipeline.types import RawAnalytics
+
+
+@task
+def extract_analytics(db) -> RawAnalytics:
```
vingkan, line 7: How about we have the `extract_analytics()` task return a `List[AnalyticsEvent]` instead of `RawAnalytics`, which is a dictionary? The dictionary is just how the data is stored in the database, but I think a list of events would be more useful to pipeline code.

We could create a new file `pipeline/types/analytics` and add an `AnalyticsEvent` type to that.

vingkan (changes requested):
Thanks for doing this @arushirai1! I left some feedback, but this is on the right track.

pipeline/extract/analytics.py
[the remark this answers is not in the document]
arushirai1: Yes, a List of analytics events sounds good!

pipeline/extract/analytics_test.py
arushirai1: Ah yes, good catch. That's right it's not separated by communities!

pipeline/extract/analytics.py
arushirai1: Okay I assumed AnalyticsEvent should be defined as a class similar to the frontend. I was wondering what the type of timestamp should be (assuming int for now).

arushirai1:
@vingkan @lwilliams717 I made changes based on your comments. Please review them!

pipeline/extract/analytics.py
```diff
@@ -0,0 +1,20 @@
...
+@task
+def extract_analytics(db) -> List[AnalyticsEvent]:
+    ref = db.reference(f"analytics/")
+    raw = ref.get()
+    list_of_analytics = []
+    for record in raw.values():
+        event = AnalyticsEvent(
+            event_type=record["event_type"] if "event_type" in record else None,
+            data=record["data"] if "data" in record else None,
+            timestamp=record["timestamp"] if "timestamp" in record else None,
```
vingkan, line 17: Instead of these ternaries you can do:

```python
record.get("event_type")
```

By default, it should return `None` if the key is not found. Or you can specify a default value:

```python
record.get("data", {})
```

pipeline/extract/analytics.py
```diff
@@ -0,0 +1,20 @@
+from typing import List
+
+from prefect import task
+
+from pipeline.types import AnalyticsEvent
+
+
+@task
+def extract_analytics(db) -> List[AnalyticsEvent]:
+    ref = db.reference(f"analytics/")
```
vingkan, line 10: No f-string needed here, nor the slash at the end.

pipeline/extract/analytics.py
```diff
@@ -0,0 +1,20 @@
...
+    raw = ref.get()
+    list_of_analytics = []
+    for record in raw.values():
+        event = AnalyticsEvent(
```
vingkan, line 14: Can you also bring in the two new values that I added in #359?

pipeline/extract/analytics.py
vingkan: It would be nice to convert the timestamp to `datetime.datetime` so that we can do more with it than just a timestamp. Check out the `NotificationInput` type for an example. When you convert the timestamp to a datetime object, you might have to divid by 1000 since it expects seconds but we store milliseconds. There might be an example somewhere else in the codebase.

pipeline/types/analytics.py
```diff
@@ -0,0 +1,9 @@
+from dataclasses import dataclass
+from typing import Any, Dict
+
+
+@dataclass
+class AnalyticsEvent:
+    event_type: str
+    data: Dict[str, Any]
+    timestamp: int
```
vingkan, line 9: I left a comment on the thread where you asked about this, but let's make timestamp a `datetime.datetime` object.

vingkan (changes requested):
Looking good! I responded to your comments and requested some changes.

pipeline/types/analytics.py
arushirai1: Sounds good! Added in latest commit.

pipeline/extract/analytics.py
arushirai1: Added!

pipeline/extract/analytics.py
arushirai1: Woops! Thanks for catching that.

pipeline/extract/analytics.py
arushirai1: xD you're right!

vingkan (approved):
Great work, thanks for doing this, Arushi! Very clean implementation and tests.

PR #361, opened 2022-08-01
Hide blocked user chats

gracesopha:
Added implementation to display a **Hidden Conversations** section on `AllChatsPage` to separate chats with blocked users. 
- Created a dictionary of blocked users and their `blocked` status. 
- Added additional `ChatInbox` to display chats with/without blocked user participants. 
- [x] Make the Hidden Conversations section a collapsible element. 
- [x] Show/Hide Hidden Conversations section based on if any chats include blocked user participants or not.

### DIFF
--- a/frontend/src/app/inbox/ChatInbox.css
+++ b/frontend/src/app/inbox/ChatInbox.css
@@ -1,3 +1,14 @@
 .ChatInbox {
     margin-top: 5px;
 }
+.HiddenInbox {
+    background: var(--color-theme-contrast);
+    color: var(--color-text);
+    box-shadow: 0 0 5px 1px var(--color-theme-accent);
+    padding: 1.5em;
+    box-sizing: border-box;
+    display: flex;
+}
+.btn {
+    margin: 15px;
+}
--- a/frontend/src/app/inbox/ChatInbox.css
+++ b/frontend/src/app/inbox/ChatInbox.css
@@ -10,5 +10,6 @@
     display: flex;
 }
 .btn {
+    cursor: pointer;
     margin: 15px;
 }
--- a/frontend/src/pages/AllChatsPage.jsx
+++ b/frontend/src/pages/AllChatsPage.jsx
@@ -1,5 +1,8 @@
-import React, { useEffect } from 'react'
+import React, { useEffect, useState } from 'react'
 import { Link } from 'react-router-dom'
+
+import { FontAwesomeIcon } from '@fortawesome/react-fontawesome'
+import { faChevronDown, faChevronUp } from '@fortawesome/free-solid-svg-icons'
 
 import { useCurrentAuthUser } from '../app/login'
 import { getUserData, useGetManyUserData, useGetAllUserMatches } from '../app/data'
@@ -45,6 +48,12 @@ export default function AllChatsPage() {
     const authUser = useCurrentAuthUser()
     const matches = useGetAllUserMatches(authUser?.uid)
     const blockedUsers = useFetchBlockedUsers(authUser?.uid)
+    const [isOpen, setIsOpen] = useState(false)
+
+    const handleInboxOpening = () => {
+        setIsOpen((prev) => !prev)
+    }
+
     // map to check if blocked user from dictionary is present in inbox
     const matchesWithBlocks = matches.map((match) => ({
         ...match,
@@ -78,10 +87,18 @@ export default function AllChatsPage() {
             <ChatInboxHeader />
             <ChatInbox chats={unblockedMatches} users={matchedUsers} />
 
-            {/* TODO: Make this a collapsable element */}
             <div className={blockedMatches.length === 0 ? 'Hidden' : ''}>
-                <h1>Hidden Conversations</h1>
-                <ChatInbox chats={blockedMatches} users={matchedUsers} />
+                <div className="Header Light">
+                    <h5>Hidden Conversations</h5>
+                    <button type="button" className="btn" onClick={handleInboxOpening}>
+                        {!isOpen ? (
+                            <FontAwesomeIcon icon={faChevronDown} />
+                        ) : (
+                            <FontAwesomeIcon icon={faChevronUp} />
+                        )}
+                    </button>
+                </div>
+                {isOpen && <ChatInbox chats={blockedMatches} users={matchedUsers} />}
             </div>
         </div>
     )
--- a/frontend/src/pages/AllChatsPage.jsx
+++ b/frontend/src/pages/AllChatsPage.jsx
@@ -46,18 +46,15 @@ export default function AllChatsPage() {
     const matches = useGetAllUserMatches(authUser?.uid)
     const blockedUsers = useFetchBlockedUsers(authUser?.uid)
 
-    const unblockedMatches = matches.filter(
-        (match) =>
-            !Object.keys(match.participants).some((participant) =>
-                Object.keys(blockedUsers).includes(participant)
-            )
-    )
+    const matchesWithBlocks = matches.map((match) => ({
+        ...match,
+        hasBlockedUser: Object.keys(match.participants).some(
+            (participant) => participant in blockedUsers
+        ),
+    }))
 
-    const blockedMatches = matches.filter((match) =>
-        Object.keys(match.participants).some((participant) =>
-            Object.keys(blockedUsers).includes(participant)
-        )
-    )
+    const unblockedMatches = matchesWithBlocks.filter((m) => !m.hasBlockedUser)
+    const blockedMatches = matchesWithBlocks.filter((m) => m.hasBlockedUser)
 
     const matchedUserIds = matches.reduce(
         (agg, m) => ({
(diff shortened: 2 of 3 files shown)


frontend/src/pages/AllChatsPage.jsx
```diff
@@ -23,16 +23,50 @@ const ChatInboxHeader = () => (
...
+
 export default function AllChatsPage() {
     const authUser = useCurrentAuthUser()
     const matches = useGetAllUserMatches(authUser?.uid)
+    const blockedUsers = useFetchBlockedUsers(authUser?.uid)
+
+    const unblockedMatches = matches.filter(
+        (match) =>
+            !Object.keys(match.participants).some((participant) =>
+                Object.keys(blockedUsers).includes(participant)
```
vingkan, line 52: Here, you can use `participant in blockedUsers` as the condition, which should check whether or not the key exists in constant time, rather than linear time to get all the keys, and then search if the key is found in the list.

frontend/src/pages/AllChatsPage.jsx
```diff
@@ -23,16 +23,50 @@ const ChatInboxHeader = () => (
...
+}
+
 export default function AllChatsPage() {
     const authUser = useCurrentAuthUser()
     const matches = useGetAllUserMatches(authUser?.uid)
+    const blockedUsers = useFetchBlockedUsers(authUser?.uid)
+
+    const unblockedMatches = matches.filter(
+        (match) =>
+            !Object.keys(match.participants).some((participant) =>
```
vingkan, line 51: This is cool, today I learned what `.some()` does! Good choice.

frontend/src/pages/AllChatsPage.jsx
```diff
@@ -45,7 +79,11 @@ export default function AllChatsPage() {
     return (
         <div className="AllChatsPage FullHeight LightBackground">
             <ChatInboxHeader />
-            <ChatInbox chats={matches} users={matchedUsers} />
+            <ChatInbox chats={unblockedMatches} users={matchedUsers} />
+
+            {/* TODO: Make this a collapsable element & hide it if no blockedMatches are present */}
```
vingkan, line 84: Looks good! And yes, the collapsible element should be added in a future PR, that way the PR is focused just on that goal and integrating it here.

frontend/src/pages/AllChatsPage.jsx
```diff
@@ -45,7 +79,11 @@ export default function AllChatsPage() {
...
+            <h1>Hidden Conversations</h1>
```
vingkan, line 85: Can we only show the hidden conversations header if there are any users that are blocked? It won't make much sense if there are no hidden conversations.

frontend/src/pages/AllChatsPage.jsx
```diff
@@ -23,16 +23,50 @@ const ChatInboxHeader = () => (
...
+            !Object.keys(match.participants).some((participant) =>
+                Object.keys(blockedUsers).includes(participant)
+            )
+    )
+
+    const blockedMatches = matches.filter((match) =>
+        Object.keys(match.participants).some((participant) =>
+            Object.keys(blockedUsers).includes(participant)
+        )
+    )
```
vingkan, line 60: Optional: Could we make this code simpler? Right now, scanning the code it looks like:

```js
const unblockedMatches = blah blah blah
  blah blah blah
  blah blah blah
  blah blah blah

const blockedMatches = blah blah blah
  blah blah blah
  blah blah blah
  blah blah blah
```

Which is not bad, at least we can tell what is going on based on the variables names. But the logic for each variable has a lot going on.

Here are three ideas, which do you like better?

Option A: Add a `hasBlockedUser` field to each match object, then filter on it

```js
const matchesWithBlocks = matches.map(...)
const unblockedMatches = matches.filter(m =>! m.hasBlockedUser)
const blockedMatches = matches.filter(m => m.hasBlockedUser)
```

Option B: Create a helper function `hasBlockedUser(match)` to handle filtering

```js
const unblockedMatches = matches.filter(m =>!hasBlockedUser(m))
const blockedMatches = matches.filter(m => hasBlockedUser(m))
```

Option C: Keep the code as-is, with each array having its own, separate filtering logic

frontend/src/pages/AllChatsPage.jsx
```diff
@@ -23,16 +23,50 @@ const ChatInboxHeader = () => (
     </div>
 )
 
+// find the blocked accounts of the user
+function useFetchBlockedUsers(myUid) {
```
vingkan, line 27: For a future PR: We could move this function to `app/data/blocked.js` and then import it here. That way the function can be re-used in places where block data is needed. This could be done in a future PR, that way it can be reused both here, and in the block-users feature.

vingkan (changes requested):
Great work, Grace! This looks good, I suggested some changes for future PRs and for this PR. Let me know if you have any questions.

frontend/src/pages/AllChatsPage.jsx
sprihajha: To me, **Option B** looks best as we won't need to add additional fields into the `matches` object and we can filter out any matches with block users based on the helper function.

frontend/src/pages/AllChatsPage.jsx
```diff
@@ -23,16 +26,53 @@ const ChatInboxHeader = () => (
...
+
 export default function AllChatsPage() {
     const authUser = useCurrentAuthUser()
     const matches = useGetAllUserMatches(authUser?.uid)
+    const blockedUsers = useFetchBlockedUsers(authUser?.uid)
+    const [isOpen, setIsOpen] = useState(false)
+
+    const handleInboxOpening = () => {
+        setIsOpen((prev) => !prev)
+    }
```
vingkan, line 55: Great job adding state to support the collapsible section!

frontend/src/pages/AllChatsPage.jsx
```diff
@@ -23,16 +26,53 @@ const ChatInboxHeader = () => (
...
+    // map to check if blocked user from dictionary is present in inbox
+    const matchesWithBlocks = matches.map((match) => ({
+        ...match,
+        hasBlockedUser: Object.keys(match.participants).some(
+            (participant) => participant in blockedUsers
+        ),
+    }))
+    // separate blocked and unblocked chats
+    const unblockedMatches = matchesWithBlocks.filter((m) => !m.hasBlockedUser)
+    const blockedMatches = matchesWithBlocks.filter((m) => m.hasBlockedUser)
```
vingkan, line 66: Nice! This is easier to follow. I know the code above is a little harder to read, but that's a consequence of how we have the match data structured. Good work!

frontend/src/pages/AllChatsPage.jsx
```diff
@@ -45,7 +85,21 @@ export default function AllChatsPage() {
...
+
+            <div className={blockedMatches.length === 0 ? 'Hidden' : ''}>
+                <div className="HiddenInbox">
+                    <h5>Hidden Conversations</h5>
+                    <button type="button" className="btn" onClick={handleInboxOpening}>
+                        {!isOpen ? (
+                            <FontAwesomeIcon icon={faChevronDown} />
+                        ) : (
+                            <FontAwesomeIcon icon={faChevronUp} />
+                        )}
```
vingkan, line 98: Nicely done adding the chevron!

Another way to write this is to save the icon as a variable and then insert that variable into the layout so that the component JSX/HTML becomes easier to read:

```jsx
const icon = isOpen
  ? (<FontAwesomeIcon icon={faChevronUp} />)
  : (<FontAwesomeIcon icon={faChevronDown} />)
...
return (
    ...
    <h5>Hidden Conversations</h5>
    <button type="button" className="btn" onClick={handleInboxOpening}>
      {icon}
    </button>
    ...
)
```

No need to change it, just wanted to show an example. This is useful when the indentation level and ternaries are getting very deep and you need a way to reduce the number of elements in one block of code so that it can be easier to read.

frontend/src/app/inbox/ChatInbox.css
```diff
@@ -1,3 +1,14 @@
...
+.HiddenInbox {
+    background: var(--color-theme-contrast);
+    color: var(--color-text);
+    box-shadow: 0 0 5px 1px var(--color-theme-accent);
+    padding: 1.5em;
+    box-sizing: border-box;
+    display: flex;
+}
+.btn {
+    margin: 15px;
```
vingkan, line 13: Optional: Can you add a button hover rule for `cursor: pointer`? This is a standard way to indicate to users that something is clickable.

frontend/src/pages/AllChatsPage.jsx
```diff
@@ -45,7 +85,21 @@ export default function AllChatsPage() {
...
+                    </button>
+                </div>
+                {isOpen && <ChatInbox chats={blockedMatches} users={matchedUsers} />}
```
vingkan, line 101: Optional: Should we add a paragraph inside the hidden section that explains something like "These chats are hidden because they include users that you have blocked."

vingkan (approved):
Great work, Grace! I left two optional comments about the hover style and hidden chats explanation, but this looks good to merge.

PR #362, opened 2022-08-01
Completed average_rating script

astep2023:
The script returns a dictionary of average ratings with the keys being user ids.

### DIFF
--- a/pipeline/scripts/average_rating.py
+++ b/pipeline/scripts/average_rating.py
@@ -1,0 +1,29 @@
+import sys
+
+sys.path.insert(0, "/workspace/butterfly/backend/utils/")
+from firebase import get_db, initialize_firebase_for_api
+
+initialize_firebase_for_api()
+
+
+def average_rating():
+    community = "sds"
+    db = get_db()
+    ref = db.reference(f"ratings/match_stars/{community}/")
+    ratings = ref.get()
+    matches = list(ratings.values())
+    totals = dict()
+    count = dict()
+    averages = dict()
+    for match in matches:
+        totals[match["from_user"]] = match["value"] + totals.get(
+            match["from_user"], 0
+        )
+        count[match["from_user"]] = 1 + count.get(match["from_user"], 0)
+    keys = list(totals.keys())
+    for i in range(len(totals)):
+        averages[keys[i]] = totals[keys[i]] / count[keys[i]]
+    return averages
+
+
+average_rating()
--- a/pipeline/scripts/average_rating.py
+++ b/pipeline/scripts/average_rating.py
@@ -1,6 +1,6 @@
 import sys
 
-sys.path.insert(0, "/workspace/butterfly/backend/utils/")
+sys.path.append("./")
 
 
 def calculate_analytics(matches):
@@ -19,7 +19,7 @@ def calculate_analytics(matches):
 
 
 def average_rating():
-    from firebase import get_db, initialize_firebase_for_api
+    from backend.utils.firebase import get_db, initialize_firebase_for_api
 
     initialize_firebase_for_api()
     if len(sys.argv) > 1:
--- a/pipeline/scripts/average_rating.py
+++ b/pipeline/scripts/average_rating.py
@@ -1,17 +1,9 @@
 import sys
 
 sys.path.insert(0, "/workspace/butterfly/backend/utils/")
-from firebase import get_db, initialize_firebase_for_api
-
-initialize_firebase_for_api()
 
 
-def average_rating():
-    community = "sds"
-    db = get_db()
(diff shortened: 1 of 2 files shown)


phodonou (approved):

pipeline/scripts/average_rating.py
```diff
@@ -0,0 +1,37 @@
+import sys
+
+sys.path.insert(0, "/workspace/butterfly/backend/utils/")
```
vingkan, line 3: This path will only work on GitPod, since it assumes `butterfly/` will be inside a `workspace/` folder.

You can do this instead:

```python
import sys

sys.path.append("./")
```

This is what the other scripts in this folder do.

pipeline/scripts/average_rating.py
```diff
@@ -0,0 +1,37 @@
...
+
+
+def calculate_analytics(matches):
```
vingkan, line 6: Nice job splitting this function out! That will make it easier to test.

Optional: You can also add a type hint to tell the reader what types the inputs and outputs should be. For example:

```python
from typing import Dict, List

from pipeline.types import UserId

def get_average_user_ratings(ratings: List[Dict]) -> Dict[UserId, float]:
  ...
  return averages
```

Also optional: consider renaming the function to describe what it does, like `get_average_user_ratings()`.

pipeline/scripts/average_rating.py
```diff
@@ -0,0 +1,37 @@
...
+        )
+        count[match["from_user"]] = 1 + count.get(match["from_user"], 0)
+    keys = list(totals.keys())
+    for i in range(len(totals)):
+        averages[keys[i]] = totals[keys[i]] / count[keys[i]]
+    return averages
+
+
+def average_rating():
+    from firebase import get_db, initialize_firebase_for_api
```
vingkan, line 22: This import should go at the top of the file. I see now why you needed to add the `backend/utils` to the path.

Instead your imports should work like this:

```python
import sys

sys.path.append("./")

from pipeline.backend.firebase import get_db, initialize_firebase_for_api
```

It would be better to import from `pipeline` than `backend`, but I see that this meets your needs, so I can refactor it later.

pipeline/scripts/average_rating.py
```diff
@@ -0,0 +1,37 @@
...
+
+    initialize_firebase_for_api()
+    if len(sys.argv) > 1:
+        community = sys.argv[1]
+    else:
+        community = "sds"
```
vingkan, line 28: This works great in this case. If you are building Python scripts in the future, you might start to have more arguments and you will want to configure their defaults and make sure that they are set correctly. A library like [argparse](https://docs.python.org/3/library/argparse.html) can be a big help for that!

pipeline/scripts/average_rating.py
```diff
@@ -0,0 +1,37 @@
...
+    db = get_db()
+    ref = db.reference(f"ratings/match_stars/{community}/")
+    ratings = ref.get()
+    matches = list(ratings.values())
+    calculate_analytics(matches)
```
vingkan, line 33: Optional: Might want to add one more step to print out the average ratings for each user.

pipeline/testing/average_rating_test.py
```diff
@@ -0,0 +1,91 @@
...
+                "value": 4,
+            },
+            {
+                "community": "sds",
+                "from_user": "user1",
+                "generator": "commonLettersGenerator",
+                "match": "2022-07-25~IRbyUHik",
+                "timestamp": 1659195148137,
+                "users": ["user1", "user4"],
+                "value": 4,
```
vingkan, line 29: Optional: There is an important case where one user might submit multiple ratings for the same match. In that case, we should only keep their latest rating.

In this case, the ratings are the same, so it does not affect the average. But if a user submitted ratings of 4 and 2 for the first match, then 2 for the second match, their average would become `8 / 3`, but really it should be `4 / 2` because the first two ratings are for the same match.

You could add another function that filters out ratings from the same user and match and only keep their most recent rating for each match.

vingkan (approved):
Great work @astep2023! I left some optional comments, mostly to give you feedback and help with your future Python projects. Feel free to merge this, since Prince has already approved!

pipeline/testing/average_rating_test.py
```diff
@@ -0,0 +1,91 @@
+import sys
+import unittest
+
+from average_rating import calculate_analytics
+
+sys.path.insert(0, "/workspace/butterfly/pipeline/scripts/")
```
vingkan, line 6: This part does need to change because the tests are failing due to this import not working correctly. Change it to this and then the tests will pass and you will be allowed to merge:

```python
import unittest

from pipeline.scripts.average_rating import calculate_analytics
```

PR #365, opened 2022-08-02
Implement Extract ChatData 

dfg-13:


pipeline/flows/matching.py
```diff
@@ -90,6 +91,7 @@ def matching_flow(defaults: Dict = {}) -> Flow:
         raw_interests = extract_interests(db, param_community)
         match_stars = extract_match_stars(db, param_community)
         intent_upvotes = extract_intent_upvotes(db, param_community)
+        chat_data = extract_recent_chatdata(db, param_community)
```
vingkan, line 94: We actually do not need to add chat data to the matching pipeline. We will need to add it to the notifications pipeline that Chris is working on. For now, it's okay to use the matching pipeline to test it. Once your PR is ready, I'll show you where to add it in the notification pipeline.

pipeline/extract/chat.py
```diff
@@ -0,0 +1,105 @@
...
+                db, community=community, chatid=id
+            )
+            if ext_message:
+                logger.info(f"Extracted message: {ext_message}")
+
+            cd1 = ChatData(
+                release_timestamp=timestamp,
+                chat_match_id=id,
+                community_id="sds",
+                participants=participants,
```
vingkan, line 63: If you look at the type definition for `ChatData`, `participants` should be a dictionary of user IDs to `User` objects. But the database only stores the user IDs. That means you need to look up the `User` objects by ID.

There are a couple ways you could accomplish that:
1. Create a separate function that takes the `ChatData` that was extract and then adds in the `User` objects to `participants`
2. Pass in a list of `User` objects to this function, and then look them up by user ID to add to the `participants` dictionary, that means you take a new argument, and that list of users must have all the users you will need
3. In this extract function, also extract all the users from the community using the function that exists, and then add them to `participants`

I would not recommend #2 because it involves making extra database requests in the same function that are not necessary. #1 or #2 would be easier to test and understand.

@notedwin can give you advice on how to do this!

pipeline/extract/chat.py
```diff
@@ -0,0 +1,105 @@
...
+                release_tag=rtag,
+                title=chatdata_title,
+                metadata=parsed_md,
+                messages=[],
```
vingkan, line 67: Similar issue here, we are expecting `ChatData` to contain a list of all the `Message` objects. You did a good job writing a separate method to extract messages. I like that approach because it's easier to test and understand.

But, we do need to populate the messages on the `ChatData` objects, so how will you do it? Check out my previous comment about the `User` objects needed in `participants` for ideas and ask @notedwin for advice.

vingkan (changes requested):
Excellent start, David! Glad you were able to figure out a lot of the Prefect, Firebase, and mock testing syntax already.

I left some comments on the biggest changes we will need in order for this to work in the pipeline.

dfg-13:
Logs from the notifications_flow in Prefect.
image

pipeline/flows/matching.py
dfg-13: I deleted this line and moved `extract_recent_chatdata` over to the notifications pipeline.

pipeline/extract/chat.py
dfg-13: A separate function was added to help with modifying the participants into showing User objects. This was then used for the `extract_recent_chatdata` function.

pipeline/extract/chat.py
dfg-13: Used the extract message function to populate the messages in the `ChatData` objects.

pipeline/extract/chat.py
```diff
@@ -0,0 +1,113 @@
...
+        return []
+    logger.info(f"Extracted {len(user_chat_records)} user-chat records.")
+    ids: Set[ChatMatchId] = set()
+    chatList: List[ChatData] = []
+
+    df_users = extract_users.run(db, community)
+    users = convert_users_into_hash(df_users)
+
+    for key in user_chat_records.keys():
+        if user_chat_records[key]["id"] not in ids:
```
vingkan, line 37: It would be more readable to invert this condition and continue the loop.

Instead of:

```python
for blah in blahs:
    if condition:
        do_thing_1()
        do_thing_2()
        do_thing_3()
```

Do this:

```python
for blah in blahs:
    if not condition:
        continue
    do_thing_1()
    do_thing_2()
    do_thing_3()
```

Check the opposite of the condition and continue to the next iteration of the loop.

This makes it easier to follow the loop logic linearly. It also avoids having deeply-indented code blocks, which become harder to read and maintain.

pipeline/transform/users.py
```diff
@@ -30,3 +31,13 @@ def convert_users_from_df(df: pd.DataFrame) -> List[User]:
     field_names = [f.name for f in dataclasses.fields(User) if f.name in cols]
     user_dicts = df[field_names].to_dict(orient="records")
     return [User(**u) for u in user_dicts]
+
+
+def convert_users_into_hash(df: pd.DataFrame) -> Dict[UserId, User]:
+    cols = set(df.columns)
+    # Parse raw community membership records into dataclasses
```
vingkan, line 38: This works! No need to change anything.

In the future, we would just refactor our extract tasks to return the dataclass objects, that way we don't have to bother with converting dataframes.

pipeline/extract/chat_test.py
```diff
@@ -0,0 +1,111 @@
...
+    mock_get_msgs.assert_called_once_with()
+
+    expected = (
+        []
+    )  # expecting an empty message in the event that there is no message records
+    assert actual == expected
+
+
+# Commented it out because I would have to create an additional mock database for users thus spending more time
+# Though, when running in Prefect the logger statements for extract_recent_chatdata() meet our expectations.
```
vingkan, line 71: This is okay to leave as-is! In the future, this also why we split up our database calls into different functions, then we can combine the results separately or pass in the data we need as inputs, that way we don't have to mock too many more database elements.

pipeline/extract/chat.py
```diff
@@ -0,0 +1,113 @@
...
+    message_ref = db.reference(f"messages/{community}/{chatid}")
+    message_records = message_ref.get()
+    l_messages: List[Message] = []
+    if not message_records:
+        logger.info(f"No message records for commmnity: {community}/{chatid}")
+    if message_records:
+        logger.info(f"Extracted {len(message_records)} message records.")
+        print(message_records)
+        for message in message_records.values():
+            print(message)
```
vingkan, line 105: You can remove the print statements now.

vingkan (approved):
Good work @dfg-13, now that this is working, the notification pipeline can work on real data!

vingkan (approved):
Well done, David! Approved to merge.

Issue #314, opened 2022-07-22
Create a generator that uses machine learning to match users


notedwin:
## Goal

Create a generator that uses machine learning to match users.

## Definition of Done

- [ ] Implement a function that takes in users with intent and interest data and converts the data to a vector.
- [ ] Implement a function to find the cosine similarity between 2 vectors.
- [ ] Implement a function to find 2 users who are the most similar.
- [ ] Create a generator that uses the previous function.
- [ ] Write test cases for your function implementations.

## Tips

- For converting users into a vector look into one-hot encoding.

PR #367, opened 2022-08-03
Clustering algo

nelischalich:
#314


pipeline/matching/generators/clustering.py
```diff
@@ -0,0 +1,86 @@
+import math
+from typing import Iterator
+
+from pipeline.matching.core import MatchGenerator
+from pipeline.types import Match, MatchingInput, MatchMetadata
+
+GENERATOR_CLUSTERING = "clusteringGenerator"
+
+# TODO: Add type annotations to auxilary functions
+def cosine_similarity(v1, v2):
```
notedwin, line 10: Try adding type annotations but if the pre-commit gives you too much hassle, we can fix this together.

pipeline/matching/generators/clustering.py
```diff
@@ -0,0 +1,86 @@
...
+
+
+def closest_match(user1, users):
+    """
+    Summary:
+        loop over all other users to find match with closest aka max cosine similarity --
+        predefined function "cosine_similarity(user1[1],user2[1])"
+        think of users as an 11 dimensional point and trying to find the closest distance between two points
+    Performance: O(?)
+    """
```
notedwin, line 31: Can you add what the big O performance is?

pipeline/matching/generators/clustering.py
```diff
@@ -0,0 +1,86 @@
...
+    cos_sim = 0
+    user1_match = ""
+    for user in users:
+        if user[0] == user1[0]:
+            continue
+        if cosine_similarity(user[1], user1[1]) > cos_sim:
+            cos_sim = cosine_similarity(user[1], user1[1])
```
notedwin, line 38: Can you think of a way to make these 2 lines more efficient since we are making 2 calls to cosine_similarity?

pipeline/matching/generators/clustering.py
```diff
@@ -0,0 +1,86 @@
...
+            user1_match = user[0]
+    return user1_match
+
+
+def converts_user_data_to_vector(users_w_data):
+    """
+    Summary:
+    Performance:
```
notedwin, line 46: Add documentation and big O performance here as well.

pipeline/matching/generators/clustering.py
```diff
@@ -0,0 +1,86 @@
...
+        for intent in user.intents:
+            c = 0
+            for i in range(5, 11):
+                if intent.name == intents[c]:
+                    if (i % 2 == 1) and intent.side.SEEKING:
+                        user_vector[i] = 1
+                    if (i % 2 == 0) and intent.side.GIVING:
+                        user_vector[i] = 1
+                if i % 2 == 0:
+                    c += 1
```
notedwin, line 66: I know you did not want that many if statements so you wrote a for loop, this gave you some more python practice.  However, this is a little harder to read if you don't know what is going to add so make sure to explain this in your documentation.

notedwin:
Make those changes and we can accept! If we had more time, I would have gone through more testing!

vingkan (approved):

PR #371, opened 2022-08-04
Pass all test cases for metrics and display_metrics

nelmajdoub:


nelmajdoub:
What would cause my code to fail the backend test?

arushirai1:
> What would cause my code to fail the backend test?

Seems like the change in `MatchingMetrics` (adding your metric-specific attributes) is not backwards compatible with the older implementation of `MatchingMetrics` which is called in `metrics_test.py`. This can be solved by giving default assignments for distribution_in_final_set and  distribution_in_community in the type definition for MatchingMetrics (see line 64).

pipeline/load/display_metrics.py
```diff
@@ -34,5 +62,18 @@ def display_internal_matching_metrics(
         output.users, metrics.n_proposed_matches_per_user
     )
     matched_user_emails = render_user_emails(output.users)
+    counts_per_intent_in_matches = render_intent_distribution_in_matches(
+        metrics.distribution_in_final_set
```
arushirai1, line 66: Good job integrating with the existing MatchingMetrics module to store the metrics!

pipeline/matching/evaluation/metrics.py
```diff
@@ -51,3 +53,49 @@ def count_selection_rate_per_generator(
             t = (selected, proposed)
             generator_selection_rate[key] = t
         return generator_selection_rate
+
+    def intent_distribution_in_selected_matches(
+        self, matches: List[Match]
+    ) -> Dict[str, int]:
+        total_count: Dict[str, int] = {}
+
+        for match in matches:
```
arushirai1, line 62: Lines 64-67 and 70-73 are identical. While in matches should not contain items in match.metadata.rareIntents or match.metadata.intents if the match isn't using the intent-based match generators, to avoid duplicated effort it would be useful to check whether the value of 'match.metadata.generator` is within the accepted list of generators (`rareIntentsGenerator`, `similarIntentsGenerator`)

pipeline/transform/compute_metrics.py
```diff
@@ -14,4 +14,19 @@ def compute_internal_matching_metrics(
     metric_output = metric.count_proposed_matches_per_user(
         output.users, iter(output.internal_data.ranked_matches)
     )
-    return MatchingMetrics(n_proposed_matches_per_user=metric_output)
+    metric_match_distribution = metric.intent_distribution_in_selected_matches(
+        output.matches
+    )
+    (
+        metric_community_distribution_1,
```
arushirai1, line 21: Use more informative variable names here. This should describe what this variable is intended to store -- the seeking and giving intent community distribution.

pipeline/types/engine.py
```diff
@@ -59,5 +59,8 @@ class MatchingOutput:
 @dataclass
 class MatchingMetrics:
     # A dictionary where keys -> users and values -> integers
+    distribution_in_final_set: Dict[str, int]
+    distribution_in_community: Tuple[Dict[str, int], Dict[str, int]]
```
arushirai1, line 63: add default value (will fix the backend test error)

pipeline/types/engine.py
```diff
@@ -59,5 +59,8 @@ class MatchingOutput:
 @dataclass
 class MatchingMetrics:
     # A dictionary where keys -> users and values -> integers
+    distribution_in_final_set: Dict[str, int]
```
arushirai1, line 62: add default value

arushirai1 (changes requested):
Good job! Needs a couple fixes but otherwise well done! Thank you for your hardwork Neveen!

arushirai1 (approved):

PR #374, opened 2022-08-04
Do not send request if the valid condition is undefined.

### LINKED ISSUE #373: Hidden chats do not update correctly after blocking a user
## Summary

@sprihajha and I noticed this issue while working on #330. After blocking a user, the hidden chats in the all chats page do not update correctly.

## Steps to Reproduce

1. Open a chat that does not have any blocked users
2. Block a user
3. Click the close and back buttons in the app to go back to the all chats page
4. The chat that you blocked a user in does not get hidden, even though it should

## Root Cause

- The `useBackendFetchJson()` does not correctly check if the user ID to check blocks for is valid.
- The `isValid` parameter is `undefined`, which leads to it being replaced by the default `true` value.
- We missed this bug because the first time we used the `isValid` parameter, there were two conditions.
- @vingkan will fix this issue in `useBackendFetchJson()` function.

vingkan:
Fixes #373.

The `isValid` parameter is responsible for determining whether or not we send the request, in case some of the values used in the route or request start out as default values that are not valid.

This PR fixes a bug where `isValid` is `undefined`, so it gets incorrectly replaced with the default value of `true`.

- [x] Test that this fixes the problem with the hidden chats on the all chats page.
- [x] Test that other pages that use this function still work as expected.

### DIFF
--- a/frontend/src/app/utils/Fetch.js
+++ b/frontend/src/app/utils/Fetch.js
@@ -19,9 +19,10 @@ export async function fetchFromBackend({ route, options = {} }) {
  * deps: React state variables for hook to depend on
  * isValid: if true, send the request, otherwise, skip it
  */
-export function useRawFetchJson({ url, options = {}, deps = [], isValid = true }) {
+export function useRawFetchJson({ url, options = {}, deps = [], isValid }) {
     const [response, setResponse] = useState(null)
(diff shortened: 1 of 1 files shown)


sprihajha (approved):
Great job debugging this so quickly! The fix looks good to merge. 🚀

PR #379, opened 2022-08-05
Rating_Reminder - Stretch Project

bridgettorres:
Waiting for feedback. Branch Conflict with Main.

### DIFF
--- a/pipeline/notifications/rating_reminder.py
+++ b/pipeline/notifications/rating_reminder.py
@@ -31,6 +31,6 @@ class RatingReminderNotifier(Notifier):
                     )
                 return remind_users
             else:
-                return List([])
+                return list([])
         else:
-            return List([])
+            return list([])
--- a/pipeline/notifications/rating_reminder.py
+++ b/pipeline/notifications/rating_reminder.py
@@ -1,0 +1,36 @@
+from typing import List
+
+from pipeline.notifications.notifier import Notifier
+from pipeline.types import NotificationInfo, NotificationInput, NotificationType
+
+
+class RatingReminderNotifier(Notifier):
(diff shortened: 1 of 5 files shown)


pipeline/notifications/rating_reminder.py
```diff
@@ -0,0 +1,36 @@
...
+        min_chat: int = 3
+        numMessages = 0
+        listUsers = inp.chatdata.participants
+        chatMessages = inp.chatdata.messages
+        dateTime = inp.run_datetime
+        for messagesSent in chatMessages:
+            if messagesSent.from_user in listUsers:
+                numMessages += 1
+        if dateTime.weekday() == 5:
+            if numMessages >= min_chat:  # should be more than 3 msg bubbles
```
chunter3, line 24: I suggest we have this as one if statement instead of a nested-if statement. So, basically, **if dateTime.weekday() == 5 && numMessages >= min_chat**. Doing so also removes the need for the additional return statement on lines 33-36, where they share an identical return value anyway.

pipeline/notifications/rating_reminder.py
```diff
@@ -0,0 +1,36 @@
...
+                        NotificationInfo(
+                            recipient=user,
+                            notification_type=NotificationType.RATE_MATCH,
+                        )
+                    )
+                return remind_users
+            else:
+                return list([])
+        else:
+            return list([])
```
chunter3, line 36: Refer to my previous comment. Basically, I believe this extra return statement is unnecessary and can be removed.

chunter3 (commented):
This is a great implementation. I left two comments but overall it's straightforward and correct. Excellent Bridget. This has my approval.

PR #381, opened 2022-08-05
Add chat preview icon for messages the user has not responded to

lacosta2:
added icons that change depending on who sent the last message in chatpreview


lacosta2:
test

frontend/src/app/inbox/ChatInbox.tsx
```diff
@@ -21,7 +21,7 @@ const ChatInbox = ({ chats, users }: ChatInboxProps) => {
         sortedChats.length > 0 &&
         sortedChats.map((c) => {
             const com = c?.communityConfig || {}
-            return <ChatPreview key={c.id} match={c} users={users} community={com} />
+            return <ChatPreview key={c.id} match={c} users={users} community={com} chatData={c} />
```
vingkan, line 24: There is already a property for this, though it is called `match`. If you look in the property list, they both have the value `c`.

frontend/src/app/inbox/ChatPreview.tsx
```diff
@@ -25,6 +28,8 @@ export default function ChatPreview({ match, users, community }) {
     const chatTitle = getChatTitle(matchedUsers)
     const userDiscs = matchedUsers.map((u) => <UserDisc key={u.uid} user={u} />)
     const userClass = matchedUsers.length === 1 ? 'Single' : 'Group'
+    const authUser = useCurrentAuthUser()
+    const currentChat: ChatData = chatData
```
vingkan, line 32: This is just a variable rename, if you want to show the type of the arguments you can do this:

```tsx
interface ChatPreviewProps {
  match: ChatData
  users: { [userId: string]: UserData }
  community: CommunityData
}

export default function ChatPreview({ match, users, community, chatData }: ChatPreviewProps) { ... }
```

frontend/src/app/inbox/ChatPreview.tsx
```diff
@@ -36,6 +41,15 @@ export default function ChatPreview({ match, users, community }) {
...
                     <p>{formatShortDate(release_timestamp)}</p>
+                    <div>
+                        {authUser?.uid === currentChat.latestMessage?.from ? (
+                            <FontAwesomeIcon icon={faCircleCheck} className="readIcon" />
+                        ) : (
+                            <div className="readIcon">
+                                <FontAwesomeIcon icon={faMoon} />
+                            </div>
+                        )}
+                    </div>
```
vingkan, line 52: You can it easier to read the JSX/HTML by saving the icon you want to a variable and then injecting it. Like this:

```tsx
// Save each icon as a variable
const iconResponded = <FontAwesomeIcon icon={faCircleCheck} className="readIcon" />
const iconNotResponded = <FontAwesomeIcon icon={faMoon} className="readIcon" />
// Save a variable for the condition
const hasResponded = authUser?.uid === currentChat.latestMessage?.from

// Save the chosen icon based on the condition to a variable
const iconRespondedStatus = hasResponded ? iconResponded : iconNotResponded

// Insert the chosen icon into the layout
return (
  ...
  <div>{iconRespondedStatus}</div>
  ...
)
```

vingkan (changes requested):
Awesome work @lacosta2! This looks really good. I love the icons you guys picked.

Here are some screenshots I took:
<div>
Screen Shot 2022-08-05 at 8 20 41 PM
Screen Shot 2022-08-05 at 8 21 04 PM
</div>

vingkan (approved):

PR #382, opened 2022-08-06
Vocable

BoominHutJr:
- added enum class
- added file handling
- added source of word lists


backend/games/word_guesser/WordGuesser.py
```diff
@@ -0,0 +1,124 @@
...
+    guesses_file = open(
+        "backend/games/word_guesser/WordGuesser_ValidGuesses.txt"
+    )
+    SOLUTIONS = solutions_file.read()
+    SOLUTION_LIST = SOLUTIONS.splitlines()
+
+    VALID_WORDS = guesses_file.read()
+    VALID_WORDS_LIST = VALID_WORDS.splitlines()
+except:
+    print("Files not readable. Check file or file path.")
```
erikqu, line 24: We might have to use the logger here rather than a print, I'm unsure @vingkan

backend/games/word_guesser/WordGuesser.py
```diff
@@ -0,0 +1,124 @@
...
+
+
+class CorrectnessScore(IntEnum):
```
erikqu, line 27: elegant!!

erikqu (approved):
LGTM!

backend/games/word_guesser/WordGuesser.py
```diff
@@ -0,0 +1,124 @@
...
+    correct = 3
+    in_word = 2
+    not_in_word = 1
```
vingkan, line 30: This enum should still use string values, because the rest of the app is expecting to get the strings we had before.

Also, usually the enum values are all caps.

backend/games/word_guesser/WordGuesser.py
```diff
@@ -0,0 +1,124 @@
...
+# initialize valid guesses and possible solutions
+SOLUTIONS: str
+VALID_WORDS: str
+
+try:
+    solutions_file = open(
+        "backend/games/word_guesser/WordGuesser_Solutions.txt"
+    )
+    guesses_file = open(
+        "backend/games/word_guesser/WordGuesser_ValidGuesses.txt"
```
vingkan, line 16: I had to fix the file paths so that this would work in our production API, make sure to merge the latest code from `main` into your branch.

backend/games/word_guesser/WordGuesser.py
```diff
@@ -0,0 +1,124 @@
...
+        if len(guess) != 5:
+            return False
+        if not guess.isalpha():
+            return False
+        if guess not in VALID_WORDS:
+            return False
+        return True
+
+    def guess_word(self, guess: str) -> Dict:
+        if self.validate_guess(guess):
```
vingkan, line 92: Should we check if this guess has already been guessed before?

backend/games/word_guesser/WordGuesser.py
```diff
@@ -0,0 +1,124 @@
...
+            return False
+        return False
+
+    # performs all logic checks on a word to determine if it is valid
+    def validate_guess(self, guess) -> bool:
+        if len(guess) != 5:
+            return False
+        if not guess.isalpha():
+            return False
+        if guess not in VALID_WORDS:
```
vingkan, line 87: You can make `VALID_WORDS` a set for faster lookup here. And you can keep `SOLUTION_WORDS` as a list for random selection. Each should the right data structure for its purpose.

backend/games/word_guesser/WordGuesser.py
```diff
@@ -0,0 +1,124 @@
...
+        self.clear_data()
+
+        new_word_index = randint(0, len(SOLUTION_LIST) - 1)  # nosec
+        self.goal_word = SOLUTION_LIST[new_word_index]
+
+    def end_game(self) -> str:
+        output = self.check_win()
+        if output == True:
+            return "You guessed the word!! Nice Job"
+        else:
```
vingkan, line 58: When you have an early return, you don't need an else statement:

```python
if output:
  return "You guessed the word!! Nice Job"
return "You failed to guess the word in 6 guesses. better luck next time."
```

backend/games/word_guesser/WordGuesser_test.py
```diff
@@ -0,0 +1,163 @@
...
+    }
+
+    normal_word = game.validate_guess("guess")
+    assert normal_word == True
+
+    normal_word = game.validate_guess("gue1s")
+    assert normal_word == False
+
+    normal_word = game.validate_guess("guesses")
+    assert normal_word == False
```
vingkan, line 126: Nice this is a well-written test!

backend/games/word_guesser/WordGuesser_test.py
```diff
@@ -0,0 +1,163 @@
...
+
+    guessed_word = game.guess_word("guesses")
+    assert guessed_word == game.guesses
+
+    guessed_word = game.guess_word("gues1")
+    assert guessed_word == game.guesses
```
vingkan, line 132: What are these testing?

backend/games/word_guesser/WordGuesser.py
```diff
@@ -0,0 +1,124 @@
...
+        self.guesses = {"guess": "", "results": []}
+        self.game_board: List[str] = []
+        self.progress: Dict = {}
+        self.goal_word = goal
+        self.guess_count = 0
+
+    def clear_data(self):
+        self.guesses = {"guess": "", "results": []}
+        self.game_board: List[str] = []
+        self.progress: Dict = {}
```
vingkan, line 44: Can you change this to a list so that we can track the order of guesses?

vingkan (changes requested):
Good job on the revisions!

There are some conflicts with the main branch, so I would recommend merging the latest code into your branch. We can go over that on our call together, the file path readings did not work in production, so I pushed a fix for that.

Also, be careful not to change the outward behavior of your class, or you may break the other parts of the app that now depend on it.

backend/games/word_guesser/WordGuesser_test.py
BoominHutJr: we are testing for guesses with more than 5 characters and for guesses that include non-alphabetical charachters

backend/games/word_guesser/WordGuesser.py
BoominHutJr: would a list of tuples be a good replacement for a dict?

vingkan: Yeah you could do a list of tuples, where the tuples are `(guess, result)` or you can just do a list of the result dictionaries.

vingkan (approved):
Excellent! I just tried `run backend-server`, which lets us simulate the production API startup so we can make sure the file path is read correctly, and that still works.

Tests are passing and this looks good to go!

PR #387, opened 2022-08-07
PR: Rating Reminder 

bridgettorres:
I closed the other PR! If I need to add more test cases please let me know!

### DIFF
--- a/pipeline/notifications/rating_reminder.py
+++ b/pipeline/notifications/rating_reminder.py
@@ -1,0 +1,33 @@
+from typing import List
+
+from pipeline.notifications.notifier import Notifier
+from pipeline.types import NotificationInfo, NotificationInput, NotificationType
+
+
+class RatingReminderNotifier(Notifier):
+    def __init__(self):
+        pass
+
+    def get_notifications(
+        self, inp: NotificationInput
+    ) -> List[NotificationInfo]:
+        remind_users = []
+        min_chat: int = 3
+        numMessages = 0
+        listUsers = inp.chatdata.participants
+        chatMessages = inp.chatdata.messages
+        dateTime = inp.run_datetime
+        for messagesSent in chatMessages:
+            if messagesSent.from_user in listUsers:
+                numMessages += 1
+        if dateTime.weekday() == 5 and numMessages >= min_chat:
+            for uid, user in listUsers.items():
+                remind_users.append(
+                    NotificationInfo(
+                        recipient=user,
+                        notification_type=NotificationType.RATE_MATCH,
+                    )
+                )
+            return remind_users
+        else:
+            return list([])
--- a/pipeline/notifications/rating_reminder_test.py
+++ b/pipeline/notifications/rating_reminder_test.py
@@ -1,0 +1,103 @@
+from datetime import datetime
+
+import pytest
+
+from pipeline.notifications import RatingReminderNotifier
+from pipeline.types import (
+    ChatData,
+    Match,
+    Message,
+    NotificationInfo,
+    NotificationInput,
+    NotificationType,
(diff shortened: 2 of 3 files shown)


pipeline/notifications/rating_reminder.py
```diff
@@ -0,0 +1,33 @@
...
+        min_chat: int = 3
+        numMessages = 0
+        listUsers = inp.chatdata.participants
+        chatMessages = inp.chatdata.messages
+        dateTime = inp.run_datetime
+        for messagesSent in chatMessages:
+            if messagesSent.from_user in listUsers:
+                numMessages += 1
+        if dateTime.weekday() == 5 and numMessages >= min_chat:
+            for uid, user in listUsers.items():
```
vingkan, line 24: This works! But the logic is not quite correct. We should only give notifications to users who sent a certain number of messages, that way if a user never messaged the chat, we don't ask them to rate the match.

What this logic does is check if the total number of messages is above a certain number, the send a notification to all users in the chat, regardless of whether or not they participated.

We can merge this and it can be a future enhancement to change the functionality!

pipeline/notifications/rating_reminder.py
```diff
@@ -0,0 +1,33 @@
...
+                remind_users.append(
+                    NotificationInfo(
+                        recipient=user,
+                        notification_type=NotificationType.RATE_MATCH,
+                    )
+                )
+            return remind_users
+        else:
+            return list([])
```
vingkan, line 33: You can just use:

```python
return []
```

That is already an empty list, so there is no need to cast it to a list again. You had to cast to an iterable before, which was slightly different.

pipeline/notifications/rating_reminder.py
```diff
@@ -0,0 +1,33 @@
...
+        if dateTime.weekday() == 5 and numMessages >= min_chat:
+            for uid, user in listUsers.items():
+                remind_users.append(
+                    NotificationInfo(
+                        recipient=user,
+                        notification_type=NotificationType.RATE_MATCH,
+                    )
+                )
+            return remind_users
+        else:
```
vingkan, line 32: You don't need multiple return statements at all. You can just return `remind_users` because if the condition is not met, then it will remain empty.

For future cases, when you have an if block that returns, there is no need for an else block, since if that condition is met, the function will return. If the condition is not met, it will proceed.

Instead you can write:

```python
if condition:
  return value
return value_otherwise
```

pipeline/notifications/rating_reminder_test.py
```diff
@@ -0,0 +1,103 @@
...
+    )
+
+    checkNotifier = RatingReminderNotifier()
+    actual = checkNotifier.get_notifications(inp)
+
+    expected = []
+    assert actual == expected
+
+
+# test less than three msgs (it should not be consired a conversation)
```
vingkan, line 80: Great job with these tests! You covered the most important cases.

pipeline/notifications/rating_reminder.py
```diff
@@ -0,0 +1,33 @@
...
+
+class RatingReminderNotifier(Notifier):
+    def __init__(self):
+        pass
+
+    def get_notifications(
+        self, inp: NotificationInput
+    ) -> List[NotificationInfo]:
+        remind_users = []
+        min_chat: int = 3
```
vingkan, line 15: No need to change this now, but for future cases, something like this could be a parameter set in the class constructor, since it is a configurable value.

vingkan (approved):
This looks great, Bridget! I have approved it and you can merge it.

Just left some optional comments for your own learning.

erikqu (approved):

PR #389, opened 2022-08-09
Rare intents hot fix

arushirai1:
Changed first case to use match.metadata.rareIntents instead of match.metadata.intents


pipeline/matching/generators/rare_intents_test.py
```diff
@@ -47,7 +47,7 @@ def testOne():
             users={"1", "3"},
             metadata=MatchMetadata(
                 generator="rareIntentsGenerator",
-                intents=[expected_intent],
+                rareIntents=[expected_intent],
```
vingkan, line 50: Intents is the correct field. Please change any references to rareIntents to use intents. See #260 for more details.

vingkan (changes requested):

